PR #202, opened 2014-12-31
refactor(front-end modules): remove legacy modules

nikgraf:
@dmitryevseev I noticed the import/export module is something we haven't used at all. Is it good to be removed?


Ashalynd:
afaik it was here because in the beginning, we tried to generate Excel on the client side, but it was too time-consuming.

Ashalynd:
Checked the whole flow, seems to be OK without these modules.

PR #204, opened 2015-01-01
fixed Save optimization, did some refactoring on BE, fixed button status...

Ashalynd:
...es for both optimization and calibration

https://trello.com/c/kisaYeTE/372-save-optimization-does-not-do-anything

(should preferably be reviewed by someone who can check both BE and FE)

### DIFF
--- a/client/source/js/modules/analysis/optimization-ctrl.js
+++ b/client/source/js/modules/analysis/optimization-ctrl.js
@@ -9,7 +9,14 @@ define([
 
       $scope.meta = meta;
       $scope.types = angular.copy(CONFIG.GRAPH_TYPES);
-      $scope.optimizationStatus = false;
+
+      var statusEnum = {
+        NOT_RUNNING: { text: "", isActive: false },
+        RUNNING: { text: "Optimization is running", isActive: true },
+        REQUESTED_TO_STOP : { text:"Optimization is requested to stop", isActive: true }
+      };
+
+      $scope.optimizationStatus = statusEnum.NOT_RUNNING;
 
       // cache placeholder
       var cachedResponse = null;
@@ -255,7 +262,7 @@ define([
       // Keep polling for updated values after every 5 seconds till we get an error.
       // Error indicates that the model is not calibrating anymore.
             optimizationTimer = $interval(checkWorkingOptimization, 5000, 0, false);
-            $scope.optimizationStatus = 'running';
+            $scope.optimizationStatus = statusEnum.RUNNING;
           } else {
             console.log("Cannot poll for optimization now");
           }
@@ -281,7 +288,7 @@ define([
         .success(function(data) {
           // Do not cancel timer yet, if the optimization is running
           if ($scope.optimizationStatus) {
-            $scope.optimizationStatus = 'requested to stop';
+            $scope.optimizationStatus = statusEnum.REQUESTED_TO_STOP;
           }
         });
     };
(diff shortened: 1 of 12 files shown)


server/src/optima/utils.py
```diff
@@ -74,6 +74,14 @@ def project_exists(name):
     cu = current_user
     return ProjectDb.query.filter_by(user_id=cu.id, name=name).count()>0
 
+def load_project(name):
```
nikgraf: great!

client/source/js/modules/analysis/optimization-ctrl.js
```diff
@@ -9,6 +9,7 @@ define([
 
       $scope.meta = meta;
       $scope.types = angular.copy(CONFIG.GRAPH_TYPES);
+      $scope.optimizationStatus = false;
```
nikgraf: How do you feel about replacing `false` by `"not running"` & making an explicit check for "not running" in the template?

"not running" tells a bit more about whats going on and mixing booleans & strings feels a bit strange to me.

Ashalynd: "false" makes it easier to check. Otherwise I'd prefer 0-based enums, not sure if they exist in Javascript?

nikgraf: As an alternative you could use an object to describe state & related meta data like the text. Here an example:

    var status = {
      NOT_RUNNING: { text: "Optimization is running", isActive: false },
      RUNNING: { text: "", isActive: true },
      REQUESTED_TO_STOP : { text:"Optimization is requested to stop", isActive: true }
    };

    // assign the message
    $scope.optimizationStatus = status.NOT_RUNNING;

    // check if it's running in the template
    ng-disabled="optimizationStatus.isActive"

    // message in the template
    {{ optimizationStatus.text }}

Ashalynd: I guess the text for RUNNING and NOT_RUNNING should be changed ;) , but overall, it seems a better solution!

nikgraf: you are right :smile:

client/source/js/modules/analysis/optimization-ctrl.js
```diff
@@ -277,8 +279,10 @@ define([
     $scope.stopOptimization = function () {
       $http.get('/api/analysis/optimization/stop')
         .success(function(data) {
-          // Cancel timer
-          stopTimer();
+          // Do not cancel timer yet, if the optimization is running
```
nikgraf: that makes more sense now :)

client/source/js/modules/analysis/optimization-ctrl.js
Ashalynd: should I do these changes or can you?

server/src/optima/optimization.py
```diff
@@ -13,7 +13,8 @@
 from flask.ext.login import login_required
 from dbconn import db
 from async_calculate import CalculatingThread, start_or_report_calculation, cancel_calculation, check_calculation
-from utils import check_project_name, project_exists, pick_params, load_model, save_working_model, report_exception
+from utils import check_project_name, project_exists, pick_params, load_model, save_working_model, \
```
nikgraf: Pylint highlighted that pick_params & save_working_model is not used in this file. Do we need them?

Ashalynd: Nope. working model is only saved within the async thread, and pick_params was not needed in optimization.py. thanks for the reminder about pylint :-}

PR #205, opened 2015-01-01
Feature/python cleanup

nikgraf:


server/src/optima/data.py
```diff
@@ -3,51 +3,51 @@
 from flask.ext.login import login_required
 from generators.line import generatedata
```
Ashalynd: I think it would be better to state the purpose of the route here, like it was. OTOH this whole module is legacy :-}

PR #206, opened 2015-01-02
Feature/378 allow user to go back and generate new spreadsheet

mkhurramali:


client/source/js/modules/model/calibration.html
```diff
@@ -1,158 +1,173 @@
...
-            <label class="checkbox-inline">
-                <input type="checkbox" ng-disabled="!canDoFitting" ng-model="enableAutomaticCalibration">
-                Automatic calibration
-            </label>
-
-            <label class="checkbox-inline">
-                <input type="checkbox" ng-disabled="!canDoFitting" ng-model="enableManualCalibration">
-                Manual calibration
-            </label>
-            <span ng-show="!canDoFitting" class="error-hint">Run "View Simulation" in order to enable data calibration</span>
```
Ashalynd: I am not sure if we should go down this road: we might end up with a messy design if different screens will offer the same functionality.

PR #210, opened 2015-01-04
Simple WB changes

cliffckerr:
This PR includes all the changes from the World Bank's feedback that were too simple to bother making Trello cards for.


client/source/js/modules/project/create-or-edit.html
```diff
@@ -54,7 +54,7 @@
               <button type="button" class="btn" ng-click="openEditPopulationModal($event, row)">Edit</button>
             </td>
             <td>
```
Ashalynd: why do you change "type" for "role"?

PR #212, opened 2015-01-05
Feature/326 3 layout of years for variable budget

mkhurramali:
https://trello.com/c/dkvEjNnt/326-3-layout-of-years-for-variable-budget


Ashalynd:
@dmitryevseev would you like to review this one?

Ashalynd:
@mkhurramali I could not see the functionality mentioned in the card actually working (see attachment in the card from what I saw). I expected to see begin and start year which would be used to initialize years range.

Ashalynd:
Ah ok - I see that these fields already existed.

PR #216, opened 2015-01-06
Feature/301 3 export all figures option

mkhurramali:
https://trello.com/c/fvmOCljy/301-3-export-all-figures-option


server/src/optima/project.py
```diff
@@ -324,6 +324,31 @@ def exportGraph():
...
+"""
+@project.route('/exportall', methods=['POST'])
+@login_required
+@report_exception()
+def exportAllGraphs():
+    from sim.makeworkbook import OptimaMultiSheetGraphTable
+    
+    data = json.loads(request.data)
+    project_name = request.headers['Project']
+
```
Ashalynd: Can't we simply pass the data structure as parameter for OptimaMultiSheetGraphTable?

server/src/sim/makeworkbook.py
```diff
@@ -628,3 +628,34 @@ def create(self, path):
                 self.formats.write_unlocked(self.sheet, row+2, i, data)
             row+=1
         self.book.close()
+
```
Ashalynd: Can't we refactor this and the other GraphTable into one class? The GraphTable with one page looks like a specific case of OptimaMultiSheetGraphTable. That would reduce the amount of duplicated code.

mkhurramali:
@Ashalynd maybe made the changes you asked for. Maybe we can pair to review this?

server/src/optima/project.py
```diff
@@ -317,13 +317,34 @@ def exportGraph():
...
+@login_required
+@report_exception()
+def exportAllGraphs():
+    from sim.makeworkbook import OptimaGraphTable
+    
+    data = json.loads(request.data)
+    project_name = request.headers['Project']
+    name = project_name + " Optimization analyses"
+    filename = name+'.xlsx'
+    path = fullpath(filename)
```
Ashalynd: There is one issue regarding sheet names: they cannot be longer than a certain no of characters (that is why I called the single-export page GRAPH DATA). did you noticed that?

client/source/js/modules/common/save-graph-as-directive.js
```diff
@@ -257,7 +292,39 @@ define(['angular', 'jquery', 'underscore', 'saveAs', './svg-to-png'],
...
+            });
+            if(showAlert) { return scope.saySorry("Sorry, some graphs cannot be exported"); }
+
+            $http({url:'/api/project/exportall',
+                  method:'POST',
+                  data: exportables,
+                  headers: {'Content-type': 'application/json'},
+                  responseType:'arraybuffer'})
+              .success(function (response, status, headers, config) {
+                var blob = new Blob([response], { type: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet' });
```
Ashalynd: There are also other pages, not only Optimization, from which the graphs are exported.

TizianoPerrucci: @Ashalynd looks like graphs are exported for those:
- AnalysisOptimization - template: optimization.html
- AnalysisScenarios - template: scenarios.html
- ModelCalibration - template: calibration.html
- ModelViewCalibration - template: cost-coverage.html

server/src/optima/project.py
```diff
@@ -317,13 +317,33 @@ def exportGraph():
...
+"""
+@project.route('/exportall', methods=['POST'])
+@login_required
+@report_exception()
+def exportAllGraphs():
+    from sim.makeworkbook import OptimaGraphTable
+    
+    data = json.loads(request.data)
+    project_name = request.headers['Project']
+    name = project_name + " Optimization analyses"
```
TizianoPerrucci: Why the file name postfix is always 'Optimization analyses' ?
Shouldn't it be dependant on which page are we exporting?

server/src/sim/makeworkbook.py
```diff
@@ -602,29 +602,46 @@ def create(self, path):
...
 
 class OptimaGraphTable:
-    def __init__ (self, name, columns, verbose = 2):
+    def __init__ (self, name, sheets, verbose = 2):
         self.name = name
-        self.columns = columns
         self.verbose = verbose
+        self.sheets = sheets
 
     def create(self, path):
```
TizianoPerrucci: Why do we need and extra function `create` if the only responsibility of the OptimaGraphTable is to produce an excel file? I would suggest we just executed the excel creation in the constructor and pass an extra argument that is the base path where the excel should be stored. The base path should not contain the name, there is already an argument for that

server/src/sim/test_makeworkbook.py
```diff
@@ -46,8 +46,12 @@ def test_ref_years_range(self):
...
     def test_create_table(self):
-        table = OptimaGraphTable('Beautiful graph', [{'title':'one', 'data':[1,2,3]}, \
-            {'title':'two', 'data':[2,3,5]},{'title':'three', 'data':["a","b","c"]}])
+        sheet = [{
+            "name":"GRAPH DATA",
+            "columns":[{'title':'one', 'data':[1,2,3]}, \
+            {'title':'two', 'data':[2,3,5]},{'title':'three', 'data':["a","b","c"]}]
+        }]
+        table = OptimaGraphTable('Beautiful graph', sheet)
         path = '/tmp/test_graph.xlsx'
```
TizianoPerrucci: I find this test wrong in several ways: 1. given a sheet name I would expect to be able to find an excel file with that name. 2. the design of OptimaGraphTable forces me to invoke create that should not be necessary 3. we are only testing the existence of a file without checking the actual content

Ashalynd: I split create from constructor so that it can be output to the given path.

TizianoPerrucci: I don't think is style. If tests are hard to read it is a reflection of the actual implementation that can be improved. Another question is if it is worth to improve the implementation and then consequently the test; I can't decide on this.

Point 3 still holds any case.

client/source/js/modules/common/save-graph-as-directive.js
mkhurramali: @TizianoPerrucci Yes I added these after @Ashalynd's comment.

server/src/optima/project.py
mkhurramali: @TizianoPerrucci thanks for catching this. The simplest solution would be to just have a generic suffix.

TizianoPerrucci: @mkhurramali can't the client just pass the name?

mkhurramali: @TizianoPerrucci I have fixed this. Actually this path is for storing temporary files only. The file that is downloaded uses the correct name depending on the screen you are on.

See line 379 on  client/source/js/modules/common/save-graph-as-directive.js

server/src/sim/makeworkbook.py
mkhurramali: I think this is presuming we may eventually have more methods in this class. If we implement this in the constructor, we will not have the opportunity to expand later.

mkhurramali:
@Ashalynd @TizianoPerrucci Apart from the test case, I have done all the other changes.

client/nginx.conf.example
```diff
@@ -17,6 +17,9 @@ server {
   listen 80;
   server_name ~^(?P<subdomain>.+?)?\.?optima\.dev$;
```
Ashalynd: that means the nginx for deployment should also change. (will do that)

PR #218, opened 2015-01-07
Feature/327 implement additional plots from functional requirements

cliffckerr:
Not tested properly, but simple enough changes that eyes 3 & 4 can hopefully sort out any bugs :)


Ashalynd:
@cliffckerr please check if anything else has to be changed in FE / API due to these additional plots

PR #221, opened 2015-01-07
fix/392-cost-coverage-curve-does-not-update

dragansah:
cost-coverage.html:
 - Fix the cc chart data and options binding expressions so that
scope.$watch in the lineScatterChart directive works properly

### DIFF
--- a/client/source/js/modules/model/cost-coverage.html
+++ b/client/source/js/modules/model/cost-coverage.html
@@ -41,8 +41,8 @@
     <div class="section" ng-if="hasCostCoverResponse">
       <h1>{{ displayedProgram.name }}</h1>
       <div class="section elastic" style="width: 470px">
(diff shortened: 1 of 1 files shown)


client/source/js/modules/model/cost-coverage.html
```diff
@@ -41,8 +41,8 @@
     <div class="section" ng-if="hasCostCoverResponse">
       <h1>{{ displayedProgram.name }}</h1>
       <div class="section elastic" style="width: 470px">
-        <div class="chart-container" ng-init="graph = graphs.plotdata_cc; graph.title = displayedProgram.name">
```
Ashalynd: This might break data export from graphs, did you check that?

dragansah: Yes, It broke export data. Fixing it but in a new PR. I'll delete this one.

dragansah:
Needs more work. Closing and making a new PR when finished.

PR #222, opened 2015-01-07
fix/392-cost-coverage-curve-does-not-update

dragansah:
cost-coverage-ctrl.js and cost-coverage.html:
- Bind the cc graph to a new scope variable "ccGraph" explicitly when preparing the graph data (instead in an ng-init as previously), enabling scope.$watch to function properly (for data and options) in the lineScatterDirective

save-graph-as-directive.js:
 - Implement new variant "ccGraph"

### DIFF
--- a/client/source/js/modules/common/save-graph-as-directive.js
+++ b/client/source/js/modules/common/save-graph-as-directive.js
@@ -52,6 +52,8 @@ define(['angular', 'jquery', 'underscore', 'saveAs', './svg-to-png'],
 
                 } else if (attrs.variant == 'coGraph') {
                   target = scope.coGraph;
+                } else if (attrs.variant == 'ccGraph') {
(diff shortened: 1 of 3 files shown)


dragansah:
@Ashalynd thoughts?

nikgraf:
@dragansah looks good :+1:

I removed the ccgraph from $scope.graphs at it is not used there anymore.

Regarding the save-as-graph-directive I was thinking about using the existing data & options attribute instead of looking for graph, ccGraph or whatever we add in the future. This way we don't need to change the directive every time we add or remove a new graph type.
What do you think @dragansah & @kitze ?

dragansah:
@nikgraf Sure I was thinking the same thing. I'll take another look tomorrow.

Also. Do we delete merged branches?

PR #224, opened 2015-01-08
fix/409-optimization-is-running-despite-the-warning-message

kitze:
optimization-ctrl.js : add a warning if the required fields aren't fllled before running optimization


client/source/js/modules/analysis/optimization-ctrl.js
```diff
@@ -267,19 +267,25 @@ define([
...
-        .success(function(data, status, headers, config) {
-          if (data.status == "OK" && data.join) {
-      // Keep polling for updated values after every 5 seconds till we get an error.
-      // Error indicates that the model is not calibrating anymore.
-            optimizationTimer = $interval(checkWorkingOptimization, 5000, 0, false);
-            $scope.optimizationStatus = statusEnum.RUNNING;
-          } else {
-            console.log("Cannot poll for optimization now");
+      $scope.startOptimization = function () {
+          if($scope.yearError === true) {
```
TizianoPerrucci: @kitze Is not clear to me where the yearError is eventually evaluated, looks like noone. Am I missing anything?

TizianoPerrucci:
@kitze I would suggest a different approach that we already use in other places, and is based on form validation. You can look into templates like cost-coverage or calibration where we check if the form is valid before invoking the operation.

kitze:
I didn't implement the checking for the year and $scope.yearError, it was already there because it was used for showing a warning if some of the fields are not filled. The Trello card was for preventing to "start optimization", it wasn't for refactoring how the checking works if there's an error in the year because that code was already there, i just used it to redirect the user to the first tab and show a modal if he hasn't filled the required fields.

TizianoPerrucci:
@kitze yes I understand your point and I didn't notice the yearError was already there. Although I still don't like it and I would prefer form validation. @nikgraf what's you take on this one?

nikgraf:
@TizianoPerrucci I prefer form validation here as well. This allows us to reduce local state that can become quite messy with a couple of them.

@kitze I think your PR was quite fine. I think the whole optimization controller is quite a mess & I would be thankful if you can use chances like this to clean up a bit & reduce complexity whenever possible.

I made a couple of changes in order to give you an idea of what kind of clean up I mean. What's your take on these @TizianoPerrucci & @kitze ?

Ashalynd:
I am sorry peeps, I messed it up due to some changes (and renamings) for graph data export, this need to be remerged now.

nikgraf:
fixed :)

TizianoPerrucci:
nice @nikgraf

kitze:
Nice job @nikgraf but i wasn't told that i can do refactoring here and there without a permission, so i was mainly focusing on solving the problem instead of refactoring the code of the controller. If you put trello cards for which controllers and files need refactoring, implementing underscore etc. that would be great and then i can check them out.

nikgraf:
@kitze I would say optimisation is one of the controllers that needs a lot of improvement. In general I think there are improvements possible everywhere.

Creating cards would need a deeper analysis. We can have a call & walk through the code if you want.

PR #226, opened 2015-01-08
358-there-is-no-feedback-from-the-app-when-click-on-export-data-in-ff34-0

kitze:
save-graph-as-directive.js: fix saving xlsx files on ff 34


TizianoPerrucci:
It does work on Firefox 34 and still works on Chrome :)

TizianoPerrucci:
From now let's focus on Chrome and leave Firefox issues at lower priority. Thx.

PR #230, opened 2015-01-09
changes to makeccocs so it can be run without parameters

robynstuart:


robynstuart:
https://trello.com/c/YOVEwiSB/403-allow-users-to-view-scatter-data-without-needing-to-enter-parameters-to-define-curves

Ashalynd:
@robynstuart is effectname[2] not going to be used at all anymore? then we should get rid of it eventually

robynstuart:
@Ashalynd yes good point. I'll have to check to see if it's used elsewhere, but if not then we can get rid of it.

PR #231, opened 2015-01-09
391-1-data-uploaded-on-populated-even-if-no-data-have-been-uploaded

kitze:
https://trello.com/c/lRuZi8uW/391-1-data-uploaded-on-populated-even-if-no-data-have-been-uploaded


nikgraf:
@kitze good job, next time would be nice to have a backend test to test the api

PR #232, opened 2015-01-09
415-error-500-when-click-on-view-calibration-and-next-menu-items-when-there-is-no-active-project

kitze:
https://trello.com/c/5uPzH6hh/415-error-500-when-click-on-view-calibration-and-next-menu-items-when-there-is-no-active-project

### DIFF
--- a/client/source/js/modules/ui/index.js
+++ b/client/source/js/modules/ui/index.js
@@ -61,8 +61,7 @@ define([
                         'Cannot proceed'
                       );
                   }
-                },
-                showClick:true
+                }
               }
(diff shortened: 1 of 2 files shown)


server/src/optima/project.py
```diff
@@ -259,8 +260,7 @@ def getProjectList():
         # Get projects for current user
         projects = ProjectDb.query.filter_by(user_id=current_user.id)
         for project in projects:
```
Ashalynd: it should be just project.data_upload_time() (considering the changes made there)

PR #235, opened 2015-01-10
feature/420-make-the-cost-coverage-outcome-graphs-update-as-soon-as-the-parameter-values-are-changed-i-e-don-t-require-users-to-click-draw-a

dragansah:
https://trello.com/c/lDUwMMzJ/420-make-the-cost-coverage-outcome-graphs-update-as-soon-as-the-parameter-values-are-changed-i-e-don-t-require-users-to-click-draw-a

feat(graphs): Make the cost-coverate-outcome graphs update as soon as the parameter values are changed. Implemented using ng-change on the 4 text fields. Implemented a 500msec debounce using underscore. There is a ngModelOptions that debounces ng-model updates, but it's in angular versions >= 1.3. The graphs are updated only if they are initially rendered by selecting the 'Draw' button.

### DIFF
--- a/client/source/js/modules/model/cost-coverage-ctrl.js
+++ b/client/source/js/modules/model/cost-coverage-ctrl.js
@@ -56,7 +56,7 @@ define(['./module', 'underscore'], function (module, _) {
         if(!categories) { // create default categories if absent
           categories = [];
           _.times(meta.progs.long.length, function (n){
-            categories.push('Others')
+            categories.push('Others');
           });}
 
         return {
@@ -290,7 +290,7 @@ define(['./module', 'underscore'], function (module, _) {
       // validation on Cost-coverage curve plotting options
       if ( !$scope.areValidParams(model.ccparams) ){
         modalService.inform(
-          function (){ null },
+          function () {},
           'Okay',
           $scope.optionsErrorMessage,
           'Error!'
@@ -326,7 +326,7 @@ define(['./module', 'underscore'], function (module, _) {
 
     $scope.changeProgram = function() {
       $scope.hasCostCoverResponse = false;
-    }
+    };
 
     /**
      * Retrieve and update graphs based on the current plot models.
@@ -343,7 +343,7 @@ define(['./module', 'underscore'], function (module, _) {
     $scope.uploadDefault = function () {
       var message = 'Upload default cost-coverage-outcome curves will be available in a future version of Optima. We are working hard in make it happen for you!';
       modalService.inform(
-        function (){ null },
+        function () {},
         'Okay',
         message,
         'Thanks for your interest!'
@@ -389,7 +389,7 @@ define(['./module', 'underscore'], function (module, _) {
       model.effect =  effectNames[graphIndex];
       if ( !$scope.areValidParams(model.coparams) ){
         modalService.inform(
-          function (){ null },
+          function () {},
           'Okay',
           $scope.optionsErrorMessage,
           'Error!'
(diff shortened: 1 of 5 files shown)


client/source/js/modules/model/cost-coverage-ctrl.js
```diff
@@ -301,6 +302,7 @@ define(['./module', 'underscore'], function (module, _) {
      * Retrieve and update graphs based on the current plot models.
      */
     $scope.generateCurves = function () {
```
Ashalynd: Considering that we already have a property $scope.hasCostCoverResponse, do we really need this one?

nikgraf: good find, I'm going to fix it

nikgraf:
@dragansah I tried to improve the UX by using a inline message instead of opening up the modal every time a parameter is changed. The inline message is only shown if `Draw` has been clicked once for the selected program.

@Ashalynd can you check the PR again?

client/source/js/modules/model/cost-coverage-ctrl.js
```diff
@@ -254,30 +258,29 @@ define(['./module', 'underscore'], function (module, _) {
...
-      return params.every(function(item){return item;});
+      return params.every(function(item) { return item; });
     };
 
     /**
-     * Returns true if none of the elements in an array is defined or not null
+     * Returns true if all of the elements in an array not defined, null or NaN
      */
-    var hasNoElements = function(params) {
-      return !params || !params.length || !hasAllElements(params);
```
Ashalynd: This will not work for empty array (e.g. params=[])

client/source/js/modules/model/cost-coverage-ctrl.js
```diff
@@ -385,19 +387,18 @@ define(['./module', 'underscore'], function (module, _) {
...
           'Okay',
-          message,
+          $scope.optionsErrorMessage,
           'Error!'
         );
         return;
       }
 
-      if (hasNoElements(model.ccparams)) delete model.ccparams;
-      if (hasNoElements(model.coparams)) delete model.coparams;
```
Ashalynd: The intention here was to remove ccparams and coparams if they haven't been set. if ccparams=[] then checking for invalid elements will not suffice: such an array has no invalid elements.

client/source/js/modules/model/cost-coverage-ctrl.js
```diff
@@ -254,30 +258,29 @@ define(['./module', 'underscore'], function (module, _) {
...
      */
-    var hasNoElements = function(params) {
-      return !params || !params.length || !hasAllElements(params);
+    var hasOnlyInvaldEntries = function(params) {
+      return params.every(function(item) {
+        return item === undefined || item === null || typeof item === "number" && isNaN(item);
+      });
+    };
+
+    $scope.areValidParams = function (params) {
```
Ashalynd: Why should we say that the params are valid if all entries are invalid? I don't get it.

client/source/js/modules/model/cost-coverage-ctrl.js
Ashalynd: ok, it's probably not relevant for now (was originating from BE it seems)

Ashalynd:
ok, merging :)

client/source/js/modules/model/cost-coverage-ctrl.js
nikgraf: Thanks, got it now. I didn't even think of the empty array case, because model.ccparams currently can't be an empty array. Because as soon as you initialise the UI you get an array filled with NaN or null depending on the input type.

I cleaned it up & made it explicit to make sure this will work even if we change things in the controller.
https://github.com/StarterSquad/Optima/commit/ef67c4e7bedbb157122a4d70fcd8c155063811aa

What do you think?

PR #240, opened 2015-01-11
424-small-changes-so-the-project-could-be-also-used-with-node-sass-instead-of-ruby-sass

kitze:
_grid.scss is taken from the ng-seed repo (which now uses node-sass)
_button.scss has just one small change - splitting multiple extends in two lines

https://trello.com/c/dqXoN5wW/424-small-changes-so-the-project-could-be-also-used-with-node-sass-instead-of-ruby-sass

### DIFF
--- a/client/source/sass/main/base/_button.scss
+++ b/client/source/sass/main/base/_button.scss
@@ -2,7 +2,8 @@
 // ======
 
 .btn {
-  @extend .btn.__m, .btn.__blue;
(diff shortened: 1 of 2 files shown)


Ashalynd:
@dmitryevseev or @nikgraf can you review this?

nikgraf:
that's fine, eventually we can switch to node-sass too

PR #241, opened 2015-01-11
275-reflow-manual-calibration-layout

kitze:
https://trello.com/c/4gXGlv4S/275-reflow-manual-calibration-layout


Ashalynd:
@kitze please add test scenario to the card.

PR #246, opened 2015-01-12
Feature/301 3 export all figures option 2

mkhurramali:
https://trello.com/c/fvmOCljy/301-3-export-all-figures-option


nikgraf:
@mkhurramali excellent work!

There was one minor hick up that I fixed on the spot. Graphs on different pages have different sizes -> made it to be dynamic.

I also decided to use 1 graph per page (landscape). No really better, but I tried to scale up the the graphs a bit.

PR #247, opened 2015-01-12
Feature/288 1 list sharing rate by population

Ashalynd:
https://trello.com/c/8sdkneJj/288-1-list-sharing-rate-by-population


Ashalynd:
Merging this, no changes in UI or API, only spreadsheet + some SIM logic.

PR #248, opened 2015-01-12
Feature/327 3 implement additional plots from functional requirements

nikgraf:
https://trello.com/c/ZNhXlABl/327-3-implement-additional-plots-from-functional-requirements


nikgraf:
@Ashalynd all export should work now:

* data for stacked charts
* data for cost coverage

I changed the colors for the stacked area graphs as well.

client/source/js/modules/analysis/optimization-ctrl.js
```diff
@@ -120,15 +113,6 @@ define(['./module', 'angular', 'd3'], function (module, angular, d3) {
     };
 
     /*
-    * Returns an array containing arrays with [x, y] for d3 line data.
```
nikgraf: We simply use _.zip now

client/source/js/modules/analysis/optimization-ctrl.js
```diff
@@ -120,15 +113,6 @@ define(['./module', 'angular', 'd3'], function (module, angular, d3) {
...
-    */
-    var generateLineData = function(xData, yData) {
```
nikgraf: Replaced it with _.zip function as it provides exactly the same result

Ashalynd:
I have found one issue: stacked charts could become misaligned. I fixed it by forcing all grid cells to be higher in we have at least one stacked chart. Not the best solution, but better than gaps.

nikgraf:
@Ashalynd ahh, that might happen in bigger screens. sorry, forget about this case

PR #250, opened 2015-01-13
feature/434-hide-stacked-option-for-scenarios-optimization-page

kitze:
https://trello.com/c/HLPIlLrI/434-hide-stacked-option-for-scenarios-optimization-page


nikgraf:
works like a charm

PR #251, opened 2015-01-13
feature/318-3-add-additional-feedback-to-optimization-screen

kitze:
Adds aditional feedback when everything in "Define objectives" is ok and the user moves to "View results".
The complete scenario is described in the description here: 
https://trello.com/c/eDgDveDB/318-3-add-additional-feedback-to-optimization-screen

Also there are some new features and small fixes (i consulted with @cliffckerr about these)
- Hide year fields when switching from variable to fixed budget
- Prevent running optimization when the budget is variable and there isn't budget for every year
- Refactor "Objective(s) to minimize" html code to be generated with ng-repeat instead of repeating things over
- Prevent user from going to the third tab when the parameters in "Define objectives" aren't ok
- Show error message if there isn't budget for every year
- Show error message if the fixed budget field isn't filled

Optimization-ctrl.js and optimization.html needs some refactoring, but a separate card should be created for that.

EDIT: 

I also fixed this card on this branch because the problem was in the same files and it would be more complicated if i created another branch for it.
https://trello.com/c/IdEB9yWq/426-1-typeerror-cannot-read-property-long-of-undefined-is-shown-when-click-on-optimiaation-analysis-and-example-xlsx-is-not-uploaded


dmitryevseev:
Please omit including code formatting in the commits. It should be done separately and on develop directly, after notifying other members. Otherwise it is likely to cause unneeded merge conflicts

client/source/js/modules/analysis/optimization-ctrl.js
```diff
@@ -275,10 +277,129 @@ define(['./module', 'angular', 'd3'], function (module, angular, d3) {
       }
     };
 
+    function validateYears(){
+      return $scope.params.objectives!==undefined &&
+             $scope.params.objectives.year!==undefined &&
+             $scope.params.objectives.year.start!==undefined &&
+             $scope.params.objectives.year.end !==undefined &&
+             ($scope.params.objectives.year.start < $scope.params.objectives.year.end);
```
dmitryevseev: It allows start '' (empty string) and end with any value.
To reproduce:
1. Set start/end years,
2. Remove `startYear` value
3. Switch to tab 3
4. See the error

client/source/js/modules/analysis/optimization-ctrl.js
```diff
@@ -275,10 +277,129 @@ define(['./module', 'angular', 'd3'], function (module, angular, d3) {
...
+    }
+
+    /**
+     * If the string is undefined return empty, otherwise just return the string
+     * @param str
+     * @returns {string}
+     */
+    function strOrEmpty(str){
+      return _(str).isUndefined() ? '' : str;
```
dmitryevseev: Mixed methods of defining `no value` cases.
In the controller vast usage of `=== undefined` but here underscore is used. And overall this function is obsolete as it can be replaced with: `str = str || '';`

client/source/js/modules/analysis/optimization-ctrl.js
dmitryevseev:
```
$scope.params.objectives
&& $scope.params.objectives.year
&& $scope.params.objectives.year.start
&& $scope.params.objectives.year.end
&& ($scope.params.objectives.year.start < $scope.params.objectives.year.end)
```

client/source/js/modules/analysis/optimization-ctrl.js
```diff
@@ -275,10 +277,129 @@ define(['./module', 'angular', 'd3'], function (module, angular, d3) {
...
+        $scope.optimizationMessage+=".";
+      }
+    }
+
+    $scope.optimizationMessage = "";
+    $scope.setActiveTab = function(tabNum){
+      if(tabNum === 3){
+        /* Prevent going to third tab if something is invalid in the first tab.
+        *  Cannot just use $scope.OptimizationForm.$invalid for this because the validation of the years and the budgets is done in a different way. */
+        if(validateYears()===false || $scope.OptimizationForm.$invalid){
```
dmitryevseev: `!validateYears()`

client/source/js/modules/analysis/optimization.html
```diff
@@ -1,540 +1,507 @@
 <div class="page">
```
dmitryevseev: Please use spaces instead of tabs - as spaces are already used all over the project

client/source/js/modules/analysis/optimization.html
```diff
@@ -1,370 +1,368 @@
...
+  <p ng-show="needData===true" class="error-hint">Please upload Optima spreadsheet first.</p>
```
dmitryevseev: ng-show already expects condition to be truthy, just ng-show="needData" is enough and verbose

dmitryevseev:
@kitze Please take a look at my comments, if something is not clear feel free to ping me in slack

client/source/js/modules/analysis/optimization-ctrl.js
kitze: I know it can be shortened with !, but it's better practice to use "===" when checking for true/false values.

client/source/js/modules/analysis/optimization.html
kitze: Same comment as above, it's better to check for true/false values with "===" especially in angular templates because angular uses different kind of javascript parser and it's clearer this way. When i paired with @nikgraf on skype he also said it's better to do explicit checks like this instead of just ng-show="needData".

client/source/js/modules/analysis/optimization-ctrl.js
kitze: I've used 
something || '' 
in other projects before and believe me that can cause more conflicts and headaches than explicit checking if something is undefined or not (also, the underscore function checks if the object equals "void 0", which is a better way to check for undefined.

client/source/js/modules/analysis/optimization-ctrl.js
kitze: As i said, the whole controller should be refactored. Especially the validation methods because there are multiple functions that basically do the same checking over and over again. I'll add a new commit with some refactoring.

kitze:
@dmitryevseev I didn't include formatting in the html, the whole optimization html just needed to be nested under  <div ng-if="needData===false">, that's why it looks like it was formatted.

client/source/js/modules/analysis/optimization-ctrl.js
dmitryevseev: I meant that all types of checks are mixed now in the code introduced. In one place it is `if(validateYears())` in another `if(validateYears()===false`. I'd like it to be consistent.

dmitryevseev:
It was reformatted. I fixed that by replacing tabs symbols with spaces that were used initially. Probably you have your IDE configured to use tabs.

kitze:
@dmitryevseev oh yes, sorry, i've replaced tabs with 2 spaces but only for .js files, now i've also set it for .html files.

Maybe it wouldn't be bad if we add an editorconfig file for the project http://editorconfig.org/

Ashalynd:
Going to merge, suggest to add cards for everything what should still be fixed.

PR #258, opened 2015-01-14
Switch between 2 and 3 parameter cost coverage curves

robynstuart:
https://trello.com/c/lz2WvNRs/444-add-additional-options-for-non-hiv-and-daly-benefits-to-cost-coverage-outcome-page
https://trello.com/c/nbq3lHt1/287-add-additional-options-for-non-hiv-and-daly-benefits


Ashalynd:
Either I am going to look at this one, or whoever merged this, do not do it without going through the whole UI flow :-}

server/src/sim/makeccocs.py
```diff
@@ -19,11 +19,12 @@
 
 ## Set defaults for testing makeccocs
 default_progname = 'SBCC'
-default_ccparams = [] # [0.9, 0.28, 154000.0, 1.0]
-default_ccplot = [] # [2e6, [0, []]]
-default_coparams = [] # [0.3, 0.5, 0.7, 0.9]
```
Ashalynd: @robynstuart let's not assume 'nan' in anything that potentially comes from frontend.

robynstuart: @Ashalynd what will come from the frontend if a textbox is left blank?

Ashalynd: None. Can it be left blank? So only first 3 fields are required?

robynstuart: yes, that's the plan.

Ashalynd: OK, replacing the default values with None should not do much harm as I see.

robynstuart: No that should be fine. I didn't know how they would come from the frontend but I assumed at least they wouldn't be floats, so the rest of the logic is based on that. I can simplify it later.

PR #262, opened 2015-01-15
layout changes to cost-coverage curves

cliffckerr:
One thing to fix before it's merged: can you make the first (cost-coverage) graph smaller? I tried changing the container from 470px to 300px but I think I changed the wrong thing since the container is actually 590px wide. I'm not an FE person :)

### DIFF
--- a/client/source/js/modules/model/cost-coverage.html
+++ b/client/source/js/modules/model/cost-coverage.html
@@ -44,7 +44,7 @@
 
       <h1>{{ displayedProgram.name }}</h1>
       <div class="section elastic">
-        <div class="elastic_col __shrink" style="width: 300px">
+        <div class="elastic_col __shrink">
           <div class="chart-container">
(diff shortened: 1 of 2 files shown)


Ashalynd:
@nikgraf , @dmitryevseev @dragansah or @kitze please have a look

@cliffckerr is there a card for that?

cliffckerr:
No, just a simple change, would take longer to make a card than to do it :)

nikgraf:
I will fix the graph width

client/source/js/modules/model/cost-coverage.html
```diff
@@ -22,34 +22,13 @@
           class="txbox"
           ng-options="p as p.name group by p.category for p in programs"
           ng-change="changeProgram()"
+          ng-click="generateCurves()"
```
nikgraf: @cliffckerr what was you intention here?

Right now this would draw the existing program. Did you want to automatically run the program once switched?

cliffckerr: Yes, that's correct

PR #263, opened 2015-01-15
allow ccplot to be empty when plotting cc curve

robynstuart:
Users should not have to specify a value for the upper limit of the x axis in order to create & display the cost coverage curve. Note that this should be fixed with the BE changes here, should not require FE changes.


server/src/sim/makeccocs.py
```diff
@@ -25,8 +25,8 @@
 default_init_ccparams = []
 default_init_convertedccparams = []
 default_init_nonhivdalys = [0.0]
-default_makeplot = 0 # CK: Otherwise brings up >100 figures
-default_effect = [['sex', 'condomcas'], [u'MSM']] # D.programs[default_progname]['effects'][0]
```
Ashalynd: do you want to have figures by default?

robynstuart: no

Ashalynd: ok, let's change it back to 0 then :)

server/src/sim/makeccocs.py
```diff
@@ -25,8 +25,8 @@
...
+default_makeplot = 1 # CK: Otherwise brings up >100 figures
```
Ashalynd: This breaks web environment, because it references to D which is not in scope.

Ashalynd: Such things are better to define within the method, when you have D available.

robynstuart: sorry anna... i just changed it to test something and forgot to change it back

Ashalynd: please commit the right version to the same branch :)

Ashalynd:
@dmitryevseev as you are working on the same part now, please review the FE changes here

PR #267, opened 2015-01-16
Improvements to validation directives, small changes in validation in CCOCS

kitze:
- deletes greaterThan and replaces it with already existing moreThan directive
- improvements to error messages directive (don't show errors if a message doesn't exist for them)
- moves validation directives into a "validation" folder instead of "common"


nikgraf:
@kitze nice cleanup

PR #275, opened 2015-01-18
Feature/461 fe changes to cost coverage outcome page

robynstuart:


client/source/js/modules/model/cost-coverage.html
```diff
@@ -70,8 +67,8 @@
               </p>
               <p>
                 3. [Optional] Specify scale-up parameters:
-                <input type="number" name="scaleUpParameter" min="1" max="10" step="0.1" class="txbox __inline __xl" placeholder="e.g.{{defaultScaleUpParameter}}"  ng-model="scaleUpParameter" ng-change="updateCurves()">
-                <error-messages for="scaleUpParameter" rules="{min:1, max:10}"></error-messages>
```
Ashalynd: @robynstuart should scale-up parameter always be <1?

robynstuart: yes.

Ashalynd: @robynstuart hmmm the graphs look totally flat with such scale-up parameter, and before, I noticed that they looked better when this parameter was >1. I see that it's being sent correctly though (as ccparams[3])

robynstuart: @Ashalynd hm... they don't look flat when I run it in python with ccparams[3] <1. Or do you mean if ccparams[3]=1?

Ashalynd: @robynstuart no< with 0, 0.8, etc. E.g. for OST with this set of parameters: ccparams: [0.9, 0.6, 400000, 0.8, 0]
ccplot: [] - I get absolutely flat line.

robynstuart: @Ashalynd It does kind of look flat for OST with those parameters... but that's ok because it just means the user will need to pick different values. But what about for MSM programs with those same parameters? That doesn't look flat at all - not when I try it, anyway. How does it look to you?

Ashalynd: @robynstuart can you tell me with which parameters OST doesn't look flat for you? MSM seems to be fine.

robynstuart: @Ashalynd try something like [0.6, 0.5, 7000000.0, None, None]. The OST program is much more expensive so ccparams[2] needs to be much higher before you really see an effect

PR #278, opened 2015-01-19
Checking circumcision and transitions

ashattock:
Seems to all be working OK


Ashalynd:
There is an issue with OST curve here if it's shown after autofit. The rest seems to be fine. @robynstuart is going to have a look at it.

PR #289, opened 2015-01-19
Fix screwy financial graphs

robynstuart:
https://trello.com/c/nwifl7nB/479-financial-commitments-are-screwy
These look okay to me now.


server/src/sim/financialanalysis.py
```diff
@@ -7,8 +7,9 @@
 import copy
 from setoptions import setoptions
 from copy import deepcopy
+from utils import sanitize
 
-def financialanalysis(D, postyear=2015, S=None, yscale='abs', makeplot=False):
+def financialanalysis(D, postyear=2015.0, S=None, yscale='abs', makeplot=False):
```
Ashalynd: That will not help if financialanalysis is still called with an integer. Better to convert an argument to a float inside the function.

PR #298, opened 2015-01-20
Feature/431 2 update metadata in db after worksheet upload

mkhurramali:
https://trello.com/c/eZ9mwuD5/431-2-update-metadata-in-db-after-worksheet-upload


server/src/optima/project.py
```diff
@@ -433,6 +433,44 @@ def uploadExcel():
         D = updatedata(D, savetofile = False)
         model = model_as_dict(D)
         project.model = model
```
Ashalynd: What about program parameters?

mkhurramali: This is fixed.

Ashalynd: OK, going to have a look soon!

PR #305, opened 2015-01-23
Feature/487 add upload download project buttons

mkhurramali:
https://trello.com/c/rVaxVzsz/487-add-upload-download-project-buttons


nikgraf:
I tested it by downloading a project, changing the project & uploading the project. Then I verified that the old data is still there. Front-end & Backend code looks good. Good to be merged from my point of view

@Ashalynd please check if it's okay that I extended the allowed file extensions by json

@mkhurramali it's not a must, but would be nice to have tests for the downloading & uploading handlers in the backend

Ashalynd:
@mkhurramali @nikgraf the implementation in this card differs from the card requirements: the card asked only for the model D to be downloaded or uploaded, not for projects and populations. Also I think the upload should be a PUT rather than a POST. Last but not least the model could have already been downloaded using /api/model/parameters (and probably upload might be logical to place there too :) )

Ashalynd:
What has to be done with "our" programs and populations would be to recreate them, the same way as was done when Excel spreadsheet was uploaded.

client/source/js/modules/ui/index.js
```diff
@@ -50,19 +50,47 @@ define([
...
+                      .success(function (response, status, headers, config) {
+                        var blob = new Blob([response], { type: 'application/json' });
+                        saveAs(blob, (activeProject.name + '.json'));
+                      })
+                      .error(function (response) {});
+                  } else {
+                    showProjectNotActive();
+                  }
+                }
+              },
```
Ashalynd: The card stated that we should first alert the user that this operation will wipe all existing data. Not done.

client/source/js/modules/ui/index.js
```diff
@@ -15,7 +15,7 @@ define([
     'app.ui.menu'
   ])
```
Ashalynd: oh and by the way, the card required to add buttons into create/manage project screen, not to the side menu.

PR #309, opened 2015-01-25
Feature/500 3 add stacked column charts in optimization instead of two pie charts

nikgraf:
https://trello.com/c/M0iVvpQc/500-3-add-stacked-column-charts-in-optimization-instead-of-two-pie-charts


Ashalynd:
Made some remarks on the card.

Ashalynd:
looks awesome now :)

nikgraf:
:smiley:

PR #310, opened 2015-01-25
Additional financial results

Ashalynd:
https://trello.com/c/gbnClflU/467-additional-financial-commitment-results

@nikgraf or @dmitryevseev : BE should be working here, please have a look at FE. This changes the number of financial graphs to 6 and introduces subtypes for annual financial graphs.


client/source/js/modules/d3-charts/d3-charts-service.js
```diff
@@ -383,6 +383,8 @@ define(['./module', 'd3', 'underscore', './scale-helpers'], function (module, d3
     }
 
     function drawAxes(scales, options, axesGroup, chartSize) {
+      if (scales === undefined) return;
```
nikgraf: when is this needed?

Ashalynd: ah that was when I tried to show graphs with no data (before I disabled that radio button)

Ashalynd: disabling was better because empty graph placeholder had a different height due to a missing legend

PR #311, opened 2015-01-26
Feature/504 all date ranges should be validated

nikgraf:
https://trello.com/c/FpZzYbwb/504-all-date-ranges-should-be-validated


Ashalynd:
works :)

PR #313, opened 2015-01-27
Loadworkbook updates

cliffckerr:
Reads in economic best, low, high future growth assumptions now.


server/src/sim/makeworkbook.py
```diff
@@ -480,9 +480,9 @@ def generate_txrx(self):
         current_row = self.emit_years_block('Number of people on subsequent lines of treatment', current_row, ['Total'], row_format = OptimaFormats.GENERAL, assumption = True)
         current_row = self.emit_years_block('Treatment eligibility criterion', current_row, ['Total'], row_format = OptimaFormats.GENERAL, assumption = True)
         current_row = self.emit_ref_years_block('Percentage of people covered by pre-exposure prophylaxis', current_row, self.pop_range, row_format = OptimaFormats.PERCENTAGE, assumption = True)
-        current_row = self.emit_years_block('Number of women on PMTCT (Option B/B+)', current_row, ['Total'], row_format = OptimaFormats.GENERAL, assumption = True)
```
Ashalynd: Could we use the same rule in naming of the parameters instead of making two per one input? I.e. number (or percentage) or number (or proportion) i/o one "parameter" for number and one "parameter" for proportion. It doesn't make sense for the user to distinguish them anyway.

Ashalynd:
Everything seems to work!

PR #314, opened 2015-01-27
feat(export): guarantee correct order

nikgraf:
https://trello.com/c/1XnxlrUP/499-1-export-all-figures-does-not-export-figures-in-the-right-order
https://trello.com/c/2PrUnTk5/497-order-of-graphs-on-model-view-is-different-from-order-on-exported-pdf

Info: To convert an SVG to an image we need to leverage async functions. In some cases it took longer to generate certain graphs & others simply where added earlier to the PDF.

In this patch we leverage promises to gurarantee the correct order. Only once all images are generated we start to generate the PDF.

### DIFF
--- a/client/source/js/modules/common/export-all-charts.js
+++ b/client/source/js/modules/common/export-all-charts.js
@@ -1,9 +1,9 @@
 define(['angular', 'jquery', 'underscore', 'jsPDF', './export-helpers-service'],
-function (angular, $, _, jspdf) {
+function (angular, $, _, JsPdf) {
   'use strict';
 
   return angular.module('app.export-all-charts', ['app.common.export-helpers'])
-  .directive('exportAllCharts', function ($http, modalService, exportHelpers) {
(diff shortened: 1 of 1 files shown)


Ashalynd:
Is a definite improvement, though does not seem to be always 100% true e.g. for the optimization.

The order in optimization: 2 pie charts, 6 x prevalence, plhiv, daly, deaths, new infections, diagnoses, first-line treatment, second-line treatment, 6 x financial commitments.

The order in pdf: 6 x prevalence, plhiv, deaths, daly, new infections, first-line treatment, diagnoses, second-line treatment, then financial commitments not in the original order. (Can change from time to time). 

Similar effect (changing of order within groups) also observed on other pages. Per se it's less of an issue than everything out of order though.

Ashalynd:
Will merge it because it's still much better than before :)

nikgraf:
oh, I need to check again. I'm not sure why it's happening. The only thing I can think of is that the jQuery selector is not providing the correct order.

PR #315, opened 2015-01-27
Feature/503 titles are missing for pie charts in optimization

nikgraf:
https://trello.com/c/yF0dQsii/503-titles-are-missing-for-pie-charts-in-optimization

In addition to show the titles properly for pie, radar & stacked charts I refactored titles used in cost coverage.

Before we used chart.title & chart.options.title. chart.title wasn't used do display in the chart, but only for the data export. I replaced it with chart.options.title & chart.options.hideTitle to remove this obscurity.


Ashalynd:
seems to work, merging :)

PR #317, opened 2015-01-27
Fix/511 year range is always 2000 2015

cliffckerr:
Fairly major change to the BE that completely changes the way time is handled. One FE change as well (removing the year options in "View simulation").


Ashalynd:
flow is working. Merging. Some financial graphs look strange, but they are going to be fixed later with another PR.

PR #319, opened 2015-01-28
added functions for normalizing F

cliffckerr:


server/src/sim/updatedata.py
```diff
@@ -87,14 +89,31 @@ def makefittedpars(D, verbose=2):
...
+        unnormF.init[p] *= D.M.hivprev[p] # Multiply by initial prevalence
+        unnormF.popsize[p] *= D.M.popsize[p][0] # Multiply by initial population size
+    unnormF.dx[3] *= D.G.datayears.mean() # Multiply by mean data year
+    return unnormF
+
+
+def normalizeF(unnormF, D):
+    """ Convert from F values that can be real-world interpretable to F values where everything is 1. """
+    normF = deepcopy(unnormF)
+    for p in range(D.G.npops):
```
Ashalynd: Are you sure that you are not dividing integers by integers?

cliffckerr: These are all numpy arrays and should maintain their data types, although I suppose casting to float would make it slightly more robust.

PR #322, opened 2015-01-28
fix/505-replace-all-alerts-with-info-modals

kitze:
https://trello.com/c/6VI5dVaA/505-replace-all-alerts-with-info-modals


sebastianconcept:
Good one @kitze !

PR #324, opened 2015-01-29
Update parameter list

cliffckerr:
Use the updated list of parameters. Seems to work except bug whereby scenario values don't get filled. See card https://trello.com/c/NDlRsp2I/517-make-updated-parameters-work


Ashalynd:
seems to be working

PR #325, opened 2015-01-29
Feature/490 more plot tools 1

mkhurramali:
https://trello.com/c/8zY9bTkH/518-more-plot-tools-line-scatter-area-chart


client/source/js/modules/common/save-graph-as-directive.js
```diff
@@ -14,7 +14,10 @@ define(['angular', 'jquery', 'underscore', 'saveAs', 'jsPDF', './svg-to-png', '.
            * event handlers.
            */
           var initialize = function() {
-            var template = '<div class="chart-buttons btn-group">' +
+            var template = '<div class="chart-left-buttons btn-group">' +
```
nikgraf: @mkhurramali The name of the directive doesn't fit anymore. I would say either change it or put the other buttons in a second directive.

PR #327, opened 2015-01-30
Circumcision checks

ashattock:
It all seems to work as expected


Ashalynd:
The financial graphs in Optimization looked strange, but this is a known bug and should be fixed in another branch. merging.

PR #329, opened 2015-01-30
Feature/490 more plot tools 2

mkhurramali:
https://github.com/StarterSquad/Optima/pull/325


nikgraf:
not ready yet, pending for now as other stuff is more important

cliffckerr:
Yeah it's not working for me yet... I thought I'd see if it was ready to merge :) BTW last commit before I merged with develop was https://github.com/StarterSquad/Optima/commit/a0fd4e066be401033d460c8789bec84a23369084
scr2015feb02-3

cliffckerr:
What's the deal with this? It's not ready to merge and probably won't be for a while if ever (if we use mpld3 instead)

Ashalynd:
@nikgraf what do you think?

nikgraf:
It's tricky. There is some really excellent groundwork here I don't want to lose. Even if we switch to mpld3 I'm pretty sure we have to improve their lib & can leverage the code here.

I'm closing the PR, but please don't remove the branch. @cliffckerr @Ashalynd

PR #335, opened 2015-02-01
Uncertainty in cost coverage

cliffckerr:
Fixes a number of major problems with optimization, including optimization results applying to all time instead of the appropriate year ranges. NB I'm not quite sure why there are so goddamn many commits in this PR :)


Ashalynd:
Closing, it's already merged

PR #340, opened 2015-02-01
fix-cost-coverage-bug-with-scale-up

robynstuart:


server/src/sim/plotccocs.py
```diff
@@ -6,9 +6,9 @@
 from matplotlib.pylab import figure, plot, hold, xlabel, ylabel, title, xlim, ylim
 from printv import printv
 
-default_progname = 'FSW programs'
-default_ccparams = [0.9, 0.5, 0.6, 400000.0, None, None] #
-default_ccplot = [1e6, None, 0]
+default_progname = 'MSM programs'
+default_ccparams = [0.9, 0.5, 0.6, 400000.0, 0.5, None] #
+default_ccplot = [100, None, 0]
 default_coparams = [0.3, 0.5, 0.7, 0.9]
```
Ashalynd: Should these be all [] ?

server/src/sim/plotccocs.py
```diff
@@ -85,7 +85,8 @@ def plotcco(D, progname=default_progname, effect=default_effect, ccparams=defaul
     ylabel(plotdata['ylabel'] )
     xlim([plotdata['xlowerlim'],plotdata['xupperlim']])
     ylim([plotdata['ylowerlim'],plotdata['yupperlim']])
-    
+
```
Ashalynd: oh oh, debug code :-}

Ashalynd:
Merging this because it fixes the error. There is still an issue with per-person costs when the maximum funding is set.

PR #349, opened 2015-02-02
Example spreadsheets

cliffckerr:
Added PWID and Sudan examples and fixed a bug that stopped Sudan from working


Ashalynd:
Tests passed :)

PR #351, opened 2015-02-03
Update financial plotting

nikgraf:
https://trello.com/c/nm7YuOGH/495-2-update-financial-plotting

Ready for review. I tested it with 2 sets: the original one & the one from Sudan. You have to go through calibration, scenarios & optimizations. See the detailed demo script on the card.

In addition I clean up tickformats for the x & y axis. From now on all of them are determined based on the scale range instead setting them manually.


server/src/sim/plotccocs.py
```diff
@@ -1,16 +1,23 @@
...
 """
 from matplotlib.pylab import figure, plot, hold, xlabel, ylabel, title, xlim, ylim
 from printv import printv
 
+# TODO robyn - which one is the correct one? I had a merge conflict
+# default_progname = 'ART'
+# default_ccparams = [2.0, 0.2, 2500000.0, None, None] #
+# default_ccplot = [1e4, None, 1]
+# default_coparams = [0.3, 0.5, 0.7, 0.9]
+
```
nikgraf: @robynstuart there was a merge conflict with the default data. I took the one from develop, but wasn't sure which one is actually the correct one. Can you check this?

Ashalynd:
Seems to work! merging.

PR #353, opened 2015-02-03
Checking sudan problems

cliffckerr:
Fixed a bug with population transitions not being validated. Um, please ignore the changes to client/source/js/config.js, maybe just cherrypick the other commit :)


nikgraf:
@cliffckerr better to revert your changes for the defaults in client/source/js/config.js. It will mess up other pages

PR #357, opened 2015-02-04
some quick changes from the faster-optimization branch

cliffckerr:


server/src/sim/gatherplotdata.py
```diff
@@ -13,7 +13,7 @@
 costtitles = {'costcum':'Cumulative HIV-related financial commitments'}
 costylabels = {}
 
-def gatheruncerdata(D, R, annual=True, verbose=2):
```
Ashalynd: should maxyear be 2030 or 2050?

Ashalynd: ok, seems to work, merging

PR #358, opened 2015-02-04
Feature/539 1 print the optimization graphs for currently chosen optimization

nikgraf:
https://trello.com/c/lOwERHww/531-3-save-the-finished-optimization-automatically
https://trello.com/c/E9MLAacW/539-1-print-the-optimization-graphs-for-currently-chosen-optimization
https://trello.com/c/lcF4iOZS/543-add-optimisation-message-for-range-in-3rd-tab

- Optimisations will now be saved once they are done.
- Switching between optimisation set entries will update the charts appropriately.
- The optimisation message in the 3rd tab was broken for range.
- The save button now is deactivated until `Checking for existing optimisation` is done.

There is one UX issue: If you run optimisation entry `x1` for example & you switch to `x2` you won't notice any updates on the charts. Nevertheless you can't start a new optimisation until the running one has stopped.


server/src/sim/optimize.py
```diff
@@ -378,38 +378,35 @@ def optimize(D, objectives=None, constraints=None, maxiters=1000, timelimit=None
...
+
     ## Gather plot data
     from gatherplotdata import gatheroptimdata
-    optim = gatheroptimdata(D, result, verbose=verbose)
-    if 'optim' not in D.plot: D.plot.optim = [] # Initialize list if required
-#    D.plot.optim.append(optim) # In any case, append
-    D.plot.optim=[optim]
-    
+    plot_result = gatheroptimdata(D, result, verbose=verbose)
+    result_to_save = {'plot': [plot_result]}
```
nikgraf: This is kind of ugly, but due the circumstances I rather keep the old structure than writing a migration for stored optimisations.

server/src/optima/optimization.py
```diff
@@ -134,95 +135,70 @@ def getWorkingModel():
...
-
         reply['status']='OK'
+        reply['name'] = 'deleted'
         reply['optimizations'] = D_dict['optimizations']
         return jsonify(reply)
-    except Exception, err:
-        reply['exception'] = traceback.format_exc()
-        return jsonify(reply)
 
-@optimization.route('/revert', methods=['POST'])
```
Ashalynd: why the second slash in the route?

nikgraf: :+1: just removed the trailing slash

Ashalynd:
Saw another error after optimization was finished:

```
 TypeError: Cannot read property 'x' of undefined
    at Object.drawAxes (d3-charts-service.js?bust=v1.0.1:418)
    at module.directive.drawGraph (line-scatter-directive.js?bust=v1.0.1:116)
    at line-scatter-directive.js?bust=v1.0.1:142
    at Object.$watchCollectionAction [as fn] (angular.js?bust=v1.0.1:12128)
    at Scope.$get.Scope.$digest (angular.js?bust=v1.0.1:12251)
    at Scope.$get.Scope.$apply (angular.js?bust=v1.0.1:12516)
    at done (angular.js?bust=v1.0.1:8204)
    at completeRequest (angular.js?bust=v1.0.1:8412)
    at XMLHttpRequest.xhr.onreadystatechange (angular.js?bust=v1.0.1:8351)
```

nikgraf:
@Ashalynd this console error is due line results with only one entry. I will fix this in another PR (which we can merge before)

nikgraf:
@Ashalynd I fixed the issue you encountered in a separate PR as other screens are also affected & in case we don't merge this yet it's better to have it in.

https://github.com/StarterSquad/Optima/pull/362

Ashalynd:
works great now! merging

PR #361, opened 2015-02-05
More sex tests

cliffckerr:
Includes a fix that enables CCOCs to be used :)


Ashalynd:
looks like it works :)

PR #362, opened 2015-02-05
Feature/546 typeerror cannot read property x of undefined

nikgraf:
https://trello.com/c/6r5KX3jp/546-typeerror-cannot-read-property-x-of-undefined-is-shown-when-checking-for-existing-optimization
https://trello.com/c/RK5HmFia/545-empty-graph-frame-at-optimization-page

The browsers crashed, because flexCeil was going into an endless loop for the value 0.
In addition I made some modification to make line charts still visually appealing even if there is only one entry for a line.

screen shot 2015-02-05 at 11 51 51

### DIFF
--- a/client/source/js/modules/d3-charts/scale-helpers.js
+++ b/client/source/js/modules/d3-charts/scale-helpers.js
@@ -34,8 +34,10 @@ define(['d3'], function (d3) {
    * 2.4 -> 3
    * 0.00242 -> 0.003
    * 0.041 -> 0.05
+   * 0 -> 0
    */
   var flexCeil = function(value) {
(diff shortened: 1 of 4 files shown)


client/source/js/modules/d3-charts/scale-helpers.js
```diff
@@ -34,8 +34,10 @@ define(['d3'], function (d3) {
    * 2.4 -> 3
    * 0.00242 -> 0.003
    * 0.041 -> 0.05
+   * 0 -> 0
    */
   var flexCeil = function(value) {
```
Ashalynd: yeah good catch :)

PR #367, opened 2015-02-06
improved logic for optimization list

Ashalynd:
A number of improvements:
- when adding optimization, add it with currently defined objectives and constraints;
- notify the user if the optimization results have not been saved yet;
- show optimization results which are still in the working model;
- do not wait before checking optimization status.


nikgraf:
@Ashalynd just created an project (example in sim) & wen to the optimisations page & switched to tab 3

    Traceback (most recent call last):
      File "/Users/nikgraf/Projects/startersquad/Optima/server/src/optima/utils.py", line 40, in __report_exception
        return api_call(*args, **kwargs)
      File "/Users/nikgraf/Projects/startersquad/Optima/server/src/optima/optimization.py", line 146, in getWorkingModel
        names = [item['name'] for item in optimizations]
    TypeError: 'NoneType' object is not iterable

PR #368, opened 2015-02-06
Adding help button

cliffckerr:
NB, I don't think it quite works yet, there may be a JS error somewhere...


client/source/js/modules/help/help.html
```diff
@@ -0,0 +1,16 @@
+<div>
+    <h1>Help</h1>
+    
+    For instructions on how to use Optima, please see the user guide:
+    <br><br>
+    <a class="btn" href='https://docs.google.com/document/d/1qQK7Q0SspvpzOt4VDqrCRELFYq1hkL3GeLEkWX06OWA/edit" target="_blank"'>Open user guide</a>
```
nikgraf: @cliffckerr ups, better to close the string :smile:

PR #374, opened 2015-03-07
Feature/555 when uploading a project a new one should be created instead of updating an existing one

nikgraf:
https://trello.com/c/8lxd4jMW/555-when-uploading-a-project-a-new-one-should-be-created-instead-of-updating-an-existing-one


nikgraf:
@Ashalynd No there is a new option on the left menu. You can enter the new name & upload a previously downloaded JSON Project file. Please take a very close look on which data I use to create project. I digged some time through the model & came to the conclusion this made the most sense. Honestly I'm not sure.

I ran calibration, cost-coverage, scenarios & optimisations with the uploaded project. All worked fine.

server/src/optima/project.py
```diff
@@ -701,17 +705,12 @@ def setData(project_id):
...
-    project = load_project(project_id)
-    if project is None:
-        reply['reason']='Project %s does not exist.' % project_id
-        return jsonify(reply)
+    data = json.loads(file.read());
 
-    data = json.load(file)
-    data['G']['projectfilename'] = project.model['G']['projectfilename']
-    data['G']['workbookname'] = project.model['G']['workbookname']
-    data['G']['projectname'] = project.model['G']['projectname']
```
Ashalynd: This should work, as long as the original project was also created via Optima web application. inputpopulations and inputprograms are not set up in non-web variant, they are read from the data.

Ashalynd:
Thanks @nikgraf , seems to work well!

PR #375, opened 2015-03-10
Adding commitments

dmitryevseev:
https://trello.com/c/wG7AXBaC/566-financial-commitments


server/src/sim/financialanalysis.py
```diff
@@ -5,16 +5,17 @@
...
-from numpy import linspace, append
+from numpy import linspace, append, npv, zeros, isnan, where
 from setoptions import setoptions
 from utils import sanitize, smoothinterp
 from printv import printv
+from datetime import date
 
-def financialanalysis(D, postyear=2015, S=None, makeplot=False, verbose=2):
+
+def financialanalysis(D, postyear=2015, S=None, makeplot=False, artgrowthrate=.05, discountrate=.03, treattime=[8,1,16,3,10], cd4time=[8,8,10,8,2,2], verbose=2):
```
nikgraf: Hey @robynstuart 

I saw you use dictionaries & lists a lot as default arguments in python methods. Are you aware that these are mutable & when modified the change will change the future default argument? It's considered best practice to not use anything except numbers, strings, None for default arguments in Python as it can cause bug really really hard to find.

Here is a good article explaining way better what I mean: http://docs.python-guide.org/en/latest/writing/gotchas/

I would propose to continuously replace them with the code pattern recommended in the article.

nikgraf:
@dmitryevseev :+1:  thx for jumping in on so short notice

For calibration I moved commitments charts to the bottom to follow the same order as in the type-selector. In scenarios & optimisations this was already the case.

dmitryevseev:
@nikgraf Oh, thanks. I see what you mean

PR #381, opened 2015-03-12
Feature/show commit version

nikgraf:
https://trello.com/c/A8tCWFDA/570-show-optima-version-number-git-hash-on-help-page

### DIFF
--- a/client/source/js/modules/help/index.js
+++ b/client/source/js/modules/help/index.js
@@ -7,9 +7,9 @@ define(['angular', '../../version', 'ui.router'], function (angular, version) {
         .state('help', {
           url: '/help',
           templateUrl: 'js/modules/help/help.html',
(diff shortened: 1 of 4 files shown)


client/source/js/modules/help/index.js
```diff
@@ -6,7 +6,10 @@ define(['angular', 'ui.router'], function (angular) {
       $stateProvider
         .state('help', {
           url: '/help',
-          templateUrl: 'js/modules/help/help.html'
+          templateUrl: 'js/modules/help/help.html',
+          controller: ['$scope', function ($scope) {
```
dmitryevseev: @nikgraf You don't need to use array inline annotation as there is ngAnnotate task run before code is minified in gulp ( https://github.com/olov/ng-annotate )

PR #382, opened 2015-03-12
Feature/496 fix chart height for mixed charts in model calibration

sebastianconcept:
https://trello.com/c/vEkCYkgu/496-2-fix-chart-height-for-mixed-charts-in-model-calibration


client/source/js/modules/model/calibration-ctrl.js
```diff
@@ -7,6 +7,11 @@ define(['./module', 'angular', 'underscore'], function (module, angular, _) {
     $scope.projectInfo = info;
     $scope.canDoFitting = $scope.projectInfo.can_calibrate;
     $scope.needData = !$scope.projectInfo.has_data;
+    $scope.$on('onAfterRender', function (e){ $scope.onAfterGraphRender() });
+    $scope.tallestGraphHeight = 0;
+    $scope.renderedGraphs = 0; // counts how many Graphs are rendered (see onAfterGraphRender for more on this)
+
+    angular.kk = $scope;
```
nikgraf: is this still needed?

client/source/js/modules/model/calibration-ctrl.js
```diff
@@ -434,5 +439,32 @@ define(['./module', 'angular', 'underscore'], function (module, angular, _) {
       updateCharts($scope.parameters.cache.response);
     }, true);
 
+    $scope.reportDataEndError = function() {
+      return "End year must be more than "+ $scope.G.dataend + ".";
+    };
+
+    // Returns the value of the tallest chart.
```
nikgraf: For methods we use the jsDoc comment convention. Better to use it here as well :smile:

nikgraf:
@sebastianconcept You added the directive also to optimisation page. What was the purpose there?

client/source/js/modules/common/after-render-directive.js
```diff
@@ -0,0 +1,17 @@
+define(['angular'], function (angular) {
+  'use strict';
+
+  return angular.module('app.common.after-render', [])
+	.directive('afterRender', [ '$timeout', function($timeout) {
```
dmitryevseev: Tabs are off in this file, and please remove inline annotation syntax - ngAnnotate takes care about it before minifying.

client/source/js/modules/model/calibration-ctrl.js
sebastianconcept: What you mean still?
That's the core of the solution.

Ah you mean the `$angular.kk`  LOL no that's to remove, good catch :+1:

client/source/js/modules/model/calibration-ctrl.js
sebastianconcept: I didn't found consistency on the project about that.

No problem adopting it

sebastianconcept:
@nikgraf about the `after-render` on the optimization page..

Derelicts from a previous experiment, thanks for noticing

client/source/js/modules/common/after-render-directive.js
sebastianconcept: ok, done.

client/source/js/modules/model/calibration-ctrl.js
nikgraf: The idea is to use // for special explanations inside functions for example & jsdoc like comments for methods/modules and so on.

dmitryevseev:
@sebastianconcept @nikgraf Are there any more pages where normalization should be applied?

nikgraf:
@dmitryevseev I though it's needed on other pages as well, but right now there is no case where we mix charts with different heights.

client/source/js/modules/common/normalise-height-directive.js
```diff
@@ -0,0 +1,122 @@
...
+     * This directive normalises the height to the tallest element having the
+     * `normalise-height` attribute currently available in the DOM.
+     *
+     * In this example both elements will end up with a height of 200px:
+     *  <div style="height: 120px" normalise-height></div>
+     *  <div style="height: 200px" normalise-height></div>
+     *
+     * Important: Any element that re-draws itself & changes the height
+     * during this process must trigger a jQuery `draw` event for the new
+     * height to be taken into account.
```
sebastianconcept: `draw` is a good method name for an object but events with names that are informative have the form of a past action. 

That's because, as you probably guessed, they are trying to signal that an something (an event) happened in the past. It probably just happened 2 nano-seconds ago, but is still the past.

client/source/js/modules/common/normalise-height-directive.js
```diff
@@ -0,0 +1,122 @@
...
+          return _.max(elements, function (element) {
+            return element.height();
+          }).height();
+        }
+      };
+
+      /**
+       * Iterates over all elements & applies the new height.
+       */
+      var _normalizeElements = function () {
```
sebastianconcept: This variable has the name of an action

client/source/js/modules/common/normalise-height-directive.js
```diff
@@ -0,0 +1,122 @@
...
+    .factory('normalisingHeightStore', function() {
+      var maxHeight = 0;
+      var elements = [];
+
+      /**
+       * Returns the max height of all existing elements.
+       *
+       * In case the list is empty 0 is returned.
+       */
+      var _maxHeightOfElement = function () {
```
sebastianconcept: Two questions:

1. Why the `_`prefix?
2. Why you made `var actionBlah = function ()` what will create all those closures in every run instead of `this.actionBlah = function ()` that would do it once during initialization?

client/source/js/modules/common/normalise-height-directive.js
```diff
@@ -0,0 +1,122 @@
...
+      /**
+       * Removes an element from the list of influencing the maximum height for
+       * normalising all the height values.
+       *
+       * Usually this would be called once an element is removed from the DOM.
+       */
+      var unregisterElement = function(element) {
+        element.off( "draw", "**" );
+        elements = _(elements).filter(function(el) { return !el.is(element); });
+        if (element.height() == maxHeight) {
```
sebastianconcept: Since they are normalised in height, `element.height() == maxHeight` is always going to be true, no?

client/source/js/modules/common/normalise-height-directive.js
```diff
@@ -0,0 +1,122 @@
...
+          // since the new normalized height might be smaller the max height of
+          // all remaining elements must be retrieved
+          var newMaxHeight = _maxHeightOfElement();
```
sebastianconcept: This is repeating code from `update`

If the `unregisterElement` method and the `update` method are doing their job right, the `unregisterElement` should only do:
1. turn off the event
2. remove the element of the collection
3. call `update`

client/source/js/modules/d3-charts/stacked-area-directive.js
```diff
@@ -159,6 +159,9 @@ define(['./module', './scale-helpers', 'angular', 'underscore'], function (modul
       );
 
       d3Charts.drawTitleAndLegend(svg, options, headerGroup);
+
+      var svgHeight = parseInt($(svg[0][0]).css('height'), 10);
+      rootElement.trigger('draw', [{height: svgHeight}]);
```
sebastianconcept: With this you are coupling the `stacked-area-chart` with the `normalise-heigh` directive.

Is loosely coupled so is not that bad but, before killing it, you had the `after-render` directive that you could have use to trigger whatever event you want from the view using the chart like:  `<stackedAreaChart  ...  after-render="$emit('element-rendered')">

PR #388, opened 2015-03-18
Feature/527 optimization plotting update how pie radar charts are plotted

nikgraf:
https://trello.com/c/NShjSesw/527-optimization-plotting-update-how-pie-radar-charts-are-plotted

The radar-chart now is optimised to show an area for low, best, height data. You can see a good example under http://optima.dev/#/graphs

The type-selector was refactored to manage the graph types completely. Thanks to @dmitryevseev for point it out & pairing with me. The types now are reseted properly for the analysis page (a regression from this commit https://github.com/optimamodel/Optima/commit/7c5ed05e2db154055a8e97d59f4b35af58e93f4f)


client/source/js/modules/analysis/optimization-ctrl.js
```diff
@@ -2,15 +2,13 @@ define(['./module', 'angular', 'underscore'], function (module, angular, _) {
...
+    $interval, meta, cfpLoadingBar, CONFIG, modalService, typeSelector,
     optimizations, optimizationHelpers) {
 
       $scope.chartsForDataExport = [];
 
       $scope.meta = meta;
-      $scope.types = graphTypeFactory.types;
-      // reset graph types every time you come to this page
-      angular.extend($scope.types, angular.copy(CONFIG.GRAPH_TYPES));
+      $scope.types = typeSelector.types;
```
sebastianconcept: Resetting controller state is something too important to sneak in between code. It should have its own method.

And I'm glad you got rid of it anyway :)

client/source/js/modules/model/calibration-ctrl.js
```diff
@@ -27,9 +27,8 @@ define(['./module', 'angular', 'underscore'], function (module, angular, _) {
       $scope.projectInfo = info;
       $scope.canDoFitting = $scope.projectInfo.can_calibrate;
 
-      $scope.types = graphTypeFactory.types;
-      // reset graph types every time you come to this page
-      angular.extend($scope.types, angular.copy(CONFIG.GRAPH_TYPES));
+      $scope.types = typeSelector.types;
```
sebastianconcept: Ditto optimization-ctrl

client/source/js/modules/ui/type-selector/type-selector-ctrl.js
```diff
@@ -1,9 +1,10 @@
 define(['./module', 'underscore'], function (module, _) {
```
sebastianconcept: Missing main comment.

For every controller (or directive) it would be good practice to add on top its main comment stating *the reason for being* of that thing.

client/source/js/modules/ui/type-selector/type-selector-ctrl.js
```diff
@@ -15,11 +16,32 @@ define(['./module', 'underscore'], function (module, _) {
       }
 
       $scope.$on('$stateChangeSuccess',function() {
-        showPartInViews('typeSelector', [
```
sebastianconcept: When the reaction to an event is implemented on the spot of where the observation is made (like here is) what happens is that you make the code the opposite of intention revealing. 

When you care to name a method with a meaningful name (probably a the verb/action) what happens, due to the inverted pyramid principle, is that you make it more intention revealing, more friendly and fast to read and cheaper to maintain.

From line 19 to line 53 what action you intended for that controller?
That's the name of the method that should encapsulate that implementation (which, by the way, could have its proper comment which is not the case now)

client/source/js/modules/ui/type-selector/type-selector-service.js
```diff
@@ -0,0 +1,36 @@
+/**
+ *  A service with helper functions for calibration.
+ */
```
sebastianconcept: Here is a nice *reason for being* comment :+1: 

Wouldn't hurt to add concrete details about what this guy do like:
1. enables costs options
2. resets calibration

And future methods extending functionality can review this comment and add it as 3. 4. etc

client/source/js/modules/analysis/optimization-ctrl.js
nikgraf: Yep, an initialise method would be nice here.

Actually I'm a bit conflicted about moving the reset to the type-selector. Now it magically resets the types manually. Once you add a page you would need to update the type-selector to handle it properly. Sounds tightly coupled to update a different module to make sure this new page/controller behaves as expected.

On the other hand it feels like the type-selectors responsibility to manage the state of the type-service.

client/source/js/modules/ui/type-selector/type-selector-ctrl.js
nikgraf: :+1:

client/source/js/modules/ui/type-selector/type-selector-ctrl.js
nikgraf: Rather than creating another method I described it in the controller comment as what happens here is the core of the controllers purpose.

Does that sound good as well?

client/source/js/modules/analysis/optimization-ctrl.js
sebastianconcept: yes, I would never be afraid of doing coupling when the concerns are right.

The optimization controller knows it has a children that can `reset` something, is okay to ask for that action when matters. What is not right is that this controller implement the reset. The how should a reset happen is a concern of the guy that holds that state. Makes sense?

nikgraf: Totally :smile:

PR #389, opened 2015-03-19
Feature/573 don t auto rerun

sebastianconcept:
https://trello.com/c/ahp0Lp3R/573-don-t-auto-rerun-model-when-spreadsheet-is-uploaded

### DIFF
--- a/server/src/optima/project.py
+++ b/server/src/optima/project.py
@@ -617,11 +617,7 @@ def uploadExcel():
         D.G.inputpopulations = deepcopy(project.populations)
 
         # Is this the first time? if so then we have to run simulations
-        try:
-            D.S # the cool kids don't reRun when they already have simulations
-            shouldReRun = False
-        except AttributeError:
-            shouldReRun = True
+        shouldReRun = S in D # the cool kids don't reRun when they already have simulations
 
         D = updatedata(D, input_programs = project.programs, savetofile = False, rerun = shouldReRun)
         model = model_as_dict(D)
--- a/server/src/optima/project.py
+++ b/server/src/optima/project.py
@@ -617,13 +617,9 @@ def uploadExcel():
         D.G.inputpopulations = deepcopy(project.populations)
 
         # Is this the first time? if so then we have to run simulations
-        try:
-            D.S # the cool kids don't reRun when they already have simulations
-            shouldReRun = False
-        except AttributeError:
-            shouldReRun = True
+        should_re_run = 'S' not in D
 
-        D = updatedata(D, input_programs = project.programs, savetofile = False, rerun = shouldReRun)
+        D = updatedata(D, input_programs = project.programs, savetofile = False, rerun = should_re_run)
         model = model_as_dict(D)
         project.model = model
         #update the programs and populations based on the data
--- a/server/src/optima/project.py
+++ b/server/src/optima/project.py
@@ -615,7 +615,15 @@ def uploadExcel():
         D.G.workbookname = D.G.projectname + '.xlsx'
         D.G.inputprograms = deepcopy(project.programs)
         D.G.inputpopulations = deepcopy(project.populations)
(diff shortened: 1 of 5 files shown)


server/src/optima/project.py
```diff
@@ -615,7 +615,15 @@ def uploadExcel():
...
         D.G.inputprograms = deepcopy(project.programs)
         D.G.inputpopulations = deepcopy(project.populations)
-        D = updatedata(D, input_programs = project.programs, savetofile = False)
+
+        # Is this the first time? if so then we have to run simulations
+        try:
+            D.S # the cool kids don't reRun when they already have simulations
+            shouldReRun = False
+        except AttributeError:
+            shouldReRun = True
```
nikgraf: This is a bit obscure ;)

you simply can do `shouldReRun = not D.has_key('S')`

nikgraf: Actually the "pythonic" way would be `shouldReRun = 'S' not in D`

nikgraf: Which should be `should_re_run = 'S' not in D` :smile:

sebastianconcept: Yes that's why I've named the variable to expose intention and also commented it.

In fact @Ashalynd suggested a nice one liner similar to that and was not giving the right answer (would not pass the spec).

On the other hand, that "obscure" thing seem to be the preferred Python way as described here:

[EAFP](https://docs.python.org/2/glossary.html#term-eafp)
```
Easier to ask for forgiveness than permission. This common Python coding style assumes the existence of valid keys or attributes and catches exceptions if the assumption proves false. This clean and fast style is characterized by the presence of many try and except statements. The technique contrasts with the LBYL style common to many other languages such as C.
```
Which I understand because

[LBYL](https://docs.python.org/2/glossary.html#term-lbyl)
````
Look before you leap. This coding style explicitly tests for pre-conditions before making calls or lookups. This style contrasts with the EAFP approach and is characterized by the presence of many if statements.

In a multi-threaded environment, the LBYL approach can risk introducing a race condition between “the looking” and “the leaping”. For example, the code, if key in mapping: return mapping[key] can fail if another thread removes key from mapping after the test, but before the lookup. This issue can be solved with locks or by using the EAFP approach.
```

nikgraf: I think "Easier to ask for forgiveness than permission" shouldn't be applied to `S`. Why?

Let's take `D` for example. We expect it to be defined always. There shouldn't be a project without a model (=D). In this case we don't want to always check for D. If project exists D exists & you can operate like that. Still you never know for sure & that's why it's important to wrap the whole accessing in a try-catch (which Flask basically does with every request).

Let's have a closer look at `S`. S is not expected to be there always. In fact we know for a certain common case S won't be defined. The information is important to enable a different code path. I would say it's even not 'LBYL', because we don't make any call on S. It's just checking a state & acting on it.

Does that make sense?

sebastianconcept: Thanks for pointing out the `_` as separator in Python style. I've read now about it now and indeed is the prevailing convention.

As I've said in http://stackoverflow.com/questions/8908760/should-i-use-camel-case-or-underscores-in-python
The "improve readability" has a little problem:
>*Function names should be lowercase, with words separated by underscores as necessary to improve readability.* 
>Which is absolutely contradictory with the fact that a camel cased group of words form an unambiguous solid symbol that looks like a single object matching the idea that a method name is one thing (as opposed to several objects, words, that have to be read and later interpreted as one single group of several things). That's even worst for _ if the method name contains a reserved word in it. 

But hey... millions are doing it, right?

nikgraf: @sebastianconcept hehe, yeah I read some underscore vs camelCase & camelCase seems to be better. Indeed in python underscore is the default & all libs we are using underscore.

If you would start out with a fresh python project would you go for camelCase for the beauty or underscore to have consistent python codebase?

nikgraf:
good stuff :+1:

PR #391, opened 2015-03-20
chore(error-handling): improves handling of server errors, outputs a reason

dmitryevseev:
Server returns a reason when file upload fails, let's use it to let user know what's exactly is wrong.


nikgraf:
Works like a charm!

screen shot 2015-03-20 at 19 01 46

PR #395, opened 2015-03-20
Fix/move optimizationmessage to the template refactoring 3

mkhurramali:


client/source/js/modules/analysis/optimization-ctrl.js
```diff
@@ -683,12 +685,16 @@ define(['./module', 'angular', 'underscore'], function (module, angular, _) {
         budgetLevel = budgetLevel + " to $" + $scope.params.objectives.outcome.budgetrange.maxval;
       }
 
-      $scope.optimizationMessage = optimizationMessageTemplate({
-        checkedPrograms : joinArrayAsSentence(validateObjectivesToMinimize().checkedPrograms, 'name', true),
-        startYear: $scope.params.objectives.year.start,
-        endYear:$scope.params.objectives.year.end,
-        budgetLevel: budgetLevel
-      });
+      if ( budgetLevel && checkedPrograms && startYear && endYear ) {
```
sebastianconcept: This `if` I see al over the place should be testing methods like `$scope.hasProgramBudgetAndYear`  or even  `$scope.hasOptimizationRequisites` 

But is out of scope of the card you took @mkhurramali so don't worry for now. This is something that should be improved on this way too large controller.

sebastianconcept:
Good to merge.
Good job @mkhurramali

nikgraf:
@mkhurramali We can the refactoring beyond the good work here. As said in the card I believe the call `constructOptimizationMessage` is not even necessary.

PR #398, opened 2015-03-21
Feature/refactor optimization ctrl

nikgraf:
https://trello.com/c/LGjHJYd2/579-move-optimizationmessage-to-the-template-refactoring

### DIFF
--- a/client/source/js/modules/analysis/optimization-ctrl.js
+++ b/client/source/js/modules/analysis/optimization-ctrl.js
@@ -697,10 +697,6 @@ define(['./module', 'angular', 'underscore'], function (module, angular, _) {
       }
     };
 
-    function constructOptimizationMessage () {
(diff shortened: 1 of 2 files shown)


nikgraf:
Something like this is what I was looking for when asking for a refactoring. I think most of is is not that exciting as I just moved code from inside one function to multiple ones. The interesting commit is the last one https://github.com/optimamodel/Optima/commit/1be7610e23cbfaf2fedf87338f916dc6f5275b1d where we remove the call to construct the optimisation message & that was my goal. Right now it happened only in two places, but you needed a deeper understanding of what's happening in these two calls. Now you just see that in the last tab the message will be constructed when the template is rendered. All we actually wanted.

@mkhurramali & @sebastianconcept please have a closer look & let me know what you think? Does it make sense? If yes, how important do you think is this kind of refactoring? If no, why? :smile:

PR #401, opened 2015-03-25
removed all status = NOK

mkhurramali:
https://trello.com/c/O5CRi4UB/578-remove-nok-from-error-responses-refactoring


nikgraf:
@mkhurramali I didn't know about returning a response together with a jsonfied object. Cool stuff!

I did a bit more as originally planned:

- Remove 'status' from every request except when checking /working for automatic calibration & optimization. Status was a duplication as the status code should indicated if the request was successful or not.

- Further I replace all json.dumps that are returned as response with jsonify: http://stackoverflow.com/a/13172658/837709

nikgraf:
@Ashalynd I merged develop in here & fixed merge conflicts

I contacted Tiziano. He also remembers that the issue was that marshaling some datastructures failed. I investigated & what I can think of is that you tried to marshal a list instead of an object. Due a security concern jsonify doesn't allow it -> http://flask.pocoo.org/docs/0.10/security/#json-security

Apart from that jsonify should work fine as it also uses json.dumps under the hood ->

https://github.com/mitsuhiko/flask/blob/master/flask/json.py#L201
https://github.com/mitsuhiko/flask/blob/master/flask/json.py#L114

Ashalynd:
ok let's give it a go!

PR #404, opened 2015-03-27
Remove bunch

cliffckerr:
Anna, I haven't updated the API. The functions bunch.toDict() and bunch.fromDict() have been replaced by dataio.tojson() and dataio.fromjson()


Ashalynd:
Let's hope it works on the server too...

PR #405, opened 2015-03-28
Add optional indicators to calibration

hugobowne:
"Add optional indicators to calibration"

### DIFF
--- a/server/src/sim/autofit.py
+++ b/server/src/sim/autofit.py
@@ -48,9 +48,14 @@ def autofit(D, timelimit=None, maxiters=500, simstartyear=2000, simendyear=2015,
             prev[p].model.x = S.tvec
             prev[p].model.y = S.people[1:,p,:].sum(axis=0) / S.people[:,p,:].sum(axis=0) # This is prevalence      
                                                         
-
+        #indicators = struct()
+        #indicatorkeys = ['death', 'newtreat', 'numtest','numinfect', 'dx']
+        #datakeys = ['death', 'newtreat', 'numtest','numinfect', 'dx']
+        for key in indicatorkeys:
+            indicators[key] = struct()
         [death, newtreat, numtest, numinfect, dx] = [[struct()], [struct()], [struct()], [struct()], [struct()]]        
- 
+        
+        # Pull out other indicators data
         mismatch = 0
         allmismatches = []
         for base in [death, newtreat, numtest, numinfect, dx]:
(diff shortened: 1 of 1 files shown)


server/src/sim/autofit.py
```diff
@@ -10,83 +10,110 @@ def autofit(D, timelimit=None, maxiters=500, simstartyear=2000, simendyear=2015,
...
+            prev[p]['model'] = dict()
+            prev[p]['data']['x'], prev[p]['data']['y'] = extractdata(D['G']['datayears'], D['data']['key']['hivprev'][0][p]) # The first 0 is for "best"
+            prev[p]['model']['x'] = S['tvec']
+            prev[p]['model']['y'] = S['people'][1:,p,:].sum(axis=0) / S['people'][:,p,:].sum(axis=0) # This is prevalence
+                                                        
+        #indicators = struct()
+#        indicatorkeys = ['death', 'newtreat', 'numtest','numinfect', 'dx']
+#        #datakeys = ['death', 'newtreat', 'numtest','numinfect', 'dx']
+#        for key in indicatorkeys:
+#            indicators[key] = dict()
```
Ashalynd: @hugobowne is it intended that every one of those indicators is an array of dictionaries?

server/src/sim/autofit.py
```diff
@@ -10,83 +10,110 @@ def autofit(D, timelimit=None, maxiters=500, simstartyear=2000, simendyear=2015,
...
+                base[0]['model']['y'] = S['newtx1'].sum(axis=0) + S['newtx2'].sum(axis=0)
+            elif base == numtest:
+                base[0]['data']['x'], base[0]['data']['y'] = extractdata(D['G']['datayears'], D['data']['opt']['numtest'][0])
+                base[0]['model']['y'] = D['M']['hivtest'].sum(axis=0)*S['people'].sum(axis=0).sum(axis=0) #testing rate x population
+            elif base == numinfect:
+                base[0]['data']['x'], base[0]['data']['y'] = extractdata(D['G']['datayears'], D['data']['opt']['numinfect'][0])
+                base[0]['model']['y'] = S['inci'].sum(axis=0)
+            elif base == dx:
+                base[0]['data']['x'], base[0]['data']['y'] = extractdata(D['G']['datayears'], D['data']['opt']['numdiag'][0])
+                base[0]['model']['y'] = S['dx'].sum(axis=0)
```
Ashalynd: As far as I can see, every indicator is always an array of size 1. Was the intention to read all of the D['data']['opt'][indicator]  array?

hugobowne: @Ashalynd D['data']['opt'][indicator][0] does read the the entire array

e.g.:
>>> D['data']['opt']['numdiag'][0]
[nan, nan, 543.0, 555.0, 567.0, 612.0, 587.0, 501.0, 603.0, 609.0, 598.0, 601.0, 656.0, 632.0, nan, nan]

does this answer your question?

server/src/sim/autofit.py
hugobowne: yes, arrays of size 1

server/src/sim/autofit.py
Ashalynd: Actually that would be still an array of length 1, it just happened to read an element which is itself an array.

Ashalynd: So my point is, perhaps you could just have ``` [death, newtreat, numtest, numinfect, dx] = [dict(), dict(), dict(), dict(), dict()]    ``` if you only use D['data']['opt'][indicator][0]

hugobowne: exactly. that is what was intended. does that make sense or am I still missing your question?

hugobowne: Note: D['data']['opt'][indicator] are actually lists of length 1 and D['data']['opt'][indicator][0] does read the entire list.

hugobowne: aaah i just saw 

"So my point is, perhaps you could just have [death, newtreat, numtest, numinfect, dx] = [dict(), dict(), dict(), dict(), dict()] if you only use D['data']['opt'][indicator][0]"

yes, i think that you're  correct, that should work -- the reason i did it in this manner was i) so that it was consistent with the indicator 'prev' and ii) also so that we can generalize it easily to calibrating over indicators to pop groups as I'm doing now for Austalia.

BUT YES. 

[death, newtreat, numtest, numinfect, dx] = [dict(), dict(), dict(), dict(), dict()]

should work.

hugobowne: oh and @Ashalynd  are you going to implement this or would you like me to?

PR #406, opened 2015-03-29
feat(export): reduce quality of the combined pdf export charts to reduce...

nikgraf:
... the file size by 2/3

https://trello.com/c/8aJqszsW/595-reduce-file-size-of-generated-pdf

By reducing the image quality the file size got reduced by 2/3 without any visible changed in the PDF (not even thin lines)


dmitryevseev:
Good one

PR #407, opened 2015-03-30
Removing all hardcoding copy

robynstuart:


server/src/sim/financialanalysis.py
```diff
@@ -48,7 +47,7 @@ def financialanalysis(D, postyear=2015, S=None, rerunmodel=False, artgrowthrate=
 
     # Get most recent ART unit costs 
     progname = 'ART'
-    prognumber = D['data']['meta']['progs']['short'].index(progname)
```
Ashalynd: This does not seem to work - there is no 'name' there.

Ashalynd: Besides, with this change all "old" projects uploaded from json will not work, because they have D['programs'] as a dictionary already.

server/src/sim/makeccocs.py
```diff
@@ -12,92 +12,70 @@
...
-            ccplot[0] = upper limit for x axis
-            ccplot[1] = None if cost data is to be displayed in current prices
-                        year if cost data is to be displayed in [year]'s prices
-            ccplot[3] = 0 if we are not adjusting by pop size
-                      = 1 if we are
-    artelig: list containing the indices for the denominator or ART coverage
-
-    Output:
-    plotdata
-    storeparams
```
Ashalynd: leftover from merging

server/src/sim/makeccocs.py
```diff
@@ -12,92 +12,70 @@
...
+    if verbose>=2: print('makecc %s %s' % (progname, ccparams))
 
-    # Check that the selected program is in the program list 
-    if progname not in D['programs'].keys():
-        raise Exception('Please select one of the following programs %s' % D['programs'].keys())
 
     # Initialise output structure
     plotdata = {}
 
     # Extract basic info from data structure
```
Ashalynd: strange that the tests don't fail though

server/src/sim/financialanalysis.py
```diff
@@ -48,7 +47,7 @@ def financialanalysis(D, postyear=2015, S=None, rerunmodel=False, artgrowthrate=
 
     # Get most recent ART unit costs
```
Ashalynd: what if ART is not one of the programs?

robynstuart: @Ashalynd i have been thinking about this - it won't be an issue in practice because every country has an ART program, but at some point we will change the way that this is done - probably ART won't be treated as a program in the same way as the other programs.

server/src/sim/financialanalysis.py
robynstuart: @Ashalynd D['programs'] is a list of dictionaries now, and 'name' is set in loadworkbook.py, line 226 or so. It will be a problem if it will break all existing projects though...

Ashalynd: @robynstuart if you can think of a way of updating D['programs'] before any calculation takes place, then it should not be a problem. Currently D['programs'] is only updated during loading data from spreadsheet (first time):
```
    if 'programs' not in D:
        D['programs'] = restructureprograms(programs)
```

server/src/sim/makeccocs.py
```diff
@@ -12,132 +12,117 @@
...
+            yvalsccu = cceqn(xvalsccpop, convertedccparams[2])
             yvalscc = [yvalsccl, yvalsccm, yvalsccu]
         else:
-            yvalsccl = cc2eqn(xvalscc, ccplottingparams[0])
-            yvalsccm = cc2eqn(xvalscc, ccplottingparams[1])
-            yvalsccu = cc2eqn(xvalscc, ccplottingparams[2])
+            yvalsccl = cc2eqn(xvalsccpop, convertedccparams[0])
+            yvalsccm = cc2eqn(xvalsccpop, convertedccparams[1])
+            yvalsccu = cc2eqn(xvalsccpop, convertedccparams[2])
             yvalscc = [yvalsccl, yvalsccm, yvalsccu]
```
nikgraf: On line 130 of makeccocs.py the variable yvalscc is referenced. Since it's initialised in an if depending on reqccparams with one program it was never initialised & it got this error:

    UnboundLocalError: local variable 'yvalscc' referenced before assignment

@robynstuart what should `yvalscc` be in this case?

robynstuart: @nikgraf oops... i fixed this in a new commit

Ashalynd:
let's give it a go. tests passed on the BE. Will require data migrations to be run.

PR #408, opened 2015-03-31
Inhomogeneities

cliffckerr:
Adds inhomogeneities to the force-of-infection. NB: WILL BREAK EVERY EXISTING PROJECT! Need to have a migration script...


nikgraf:
@Ashalynd & I both tested it on our machines. Good to be merged!

PR #410, opened 2015-03-31
fix(sass.menu): enable hardware acceleration of menu animations to have ...

dmitryevseev:
...it behave nicely in webkit, fixes menu glitch on retina

https://trello.com/c/2cLjC4xp/596-fix-menu-layout-glitch-on-retina-screens


nikgraf:
@dmitryevseev wow, is there any background info? why is this happening?

PR #411, opened 2015-04-02
Drag and resize charts in model/calibration using angular-gridster

dragansah:
- Added a dependency to angular-gridster in bower.json so a rebuild should be done
 - The grid layout is 16 columns and the initial chart length is 4 columns totaling 4 charts per row
 - The chart width/height calculations are not ideal but based on existing fixed pixel offsets based on the padding of the d3 charts


dragansah:
https://trello.com/c/8NdGzA1a/598-drag-and-resize-charts-in-model-calibration-using-angular-gridster

client/source/js/modules/model/calibration-ctrl.js
```diff
@@ -63,6 +66,59 @@ define(['./module', 'angular', 'underscore'], function (module, angular, _) {
...
+      });
+      $scope.showCharts = true;
+      // shallow copy instead of deep copy with angular.copy
+      if($scope.charts.length > 0)
+        $scope.charts[0] = _.clone($scope.charts[0]);
+    };
+
+    var resizeChart = function(chart, newWidth, newHeight) {
+      var widthOffset = 20;
+      var heightOffset = chart.type == 'stackedAreaChart' ? 210 : 100;
```
dragansah: Can be improved.

nikgraf: @dragansah Why the offset & why is it different for stackedAreaCarts?

client/source/js/modules/model/calibration-ctrl.js
```diff
@@ -63,6 +66,59 @@ define(['./module', 'angular', 'underscore'], function (module, angular, _) {
...
+      chart.options.width = newWidth - widthOffset;
+      chart.options.height = newHeight - heightOffset;
+      chart.id = generateChartId();
```
nikgraf: Math.random has a change of not being unique. You can use http://underscorejs.org/#uniqueId to be sure it's unique in the DOM.

nikgraf:
@dragansah Do you think you could move the gridster functionality to one or multiple directives? It's a lot of code making the controller more complex. Right now we try so simplify them & move logic to various components. Another benefit would be that it could easily be applied to other pages as well.

What do you think?

dragansah:
@nikgraf I did it as two separate directives resizableChartPanel and resizableChart. Please review.

client/source/js/modules/model/calibration-ctrl.js
dragansah: @nikgraf Done in the latest commit. Thanks.

dragansah:
@nikgraf  The code is there but I'm having problems with the formatting of calibration.html. I'll re-apply the changes to develop in a new commit so that the diff is clean.

dragansah:
Re-applied

client/source/js/modules/ui/gridster/resizable-chart-panel-directive.js
```diff
@@ -0,0 +1,102 @@
...
+         * We use the chart id to track by id in ng-repeat to avoid redrawing of all the charts when the chart array
+         * is updated when resizing the charts.
+         * */
+        var generateChartId = function () {
+          return _.uniqueId("chart_");
+        };
+
+        var resizeChart = function (chart, newWidth, newHeight) {
+          var widthOffset = 20;
+          var heightOffset = chart.type == 'stackedAreaChart' ? 210 : 100;
```
dragansah: @nikgraf The difference in the offset is the legend + padding

nikgraf:
@dragansah good stuff, I will take a closer look tomorrow :)

nikgraf:
@dragansah sorry, I didn't get to it yet. Most likely tomorrow

client/source/js/modules/ui/gridster/resizable-chart-panel-directive.js
nikgraf: @dragansah I believe going for a fixed height is bad choice here. There can be many more populations (which result in more labels). They height should be calculated depending on the amount or in best case dynamically like the normalise-height directive does.

nikgraf:
I took a closer look. The code quality is good & I really appreciate you adding comments.

There are a couple issues I found. I think some should be fixed, but some would also need input from @cliffckerr on how he imagines the UI should work.

Demo:

1. Resize a chart to be larger
2. Run automatic calibration
3. The charts don't update properly

screen shot 2015-04-10 at 10 22 55


Demo:

1. Resize a chart to be larger
2. Update a value from manual calibration
3. The charts don't update properly

screen shot 2015-04-10 at 10 24 14


Demo:

1. Open calibration page on a Macbook Pro 13" Retina screen
2. Naturally there were 2 columns on 1280x1024 screen. It looks messy.
  -> Can we make the columns count depending on the screensize? @cliffckerr What do you think?

screen shot 2015-04-10 at 10 11 46


Demo:

1. Resize a couple charts (smaller & larger)
2. Export all charts
3. Still every chart is on page, but the size influenced the export.
  -> charts can be way to big or too small.
  I'm not sure what the expected behavior should be here. /cc @cliffckerr 
  
Demo:

1. Enable the manual calibration panel
2. Disable the manual calibration panel
3. The content should go back to fill out the full width.
   -> @cliffckerr maybe not so important?
  
I really like that the smaller charts (in height) scale up to the full height of the ones with labels.

nikgraf:
@dragansah did you have time to look into the issues?

PR #412, opened 2015-04-04
Add updated_time field to project model

dragansah:
https://trello.com/c/dDlUCAgD/599-redo-project-metadata-table
  - updated fields on projects home view
  - updated fields on project open view

### DIFF
--- a/server/db/versions/010_add_updated_time_field_postgresql_downgrade.sql
+++ b/server/db/versions/010_add_updated_time_field_postgresql_downgrade.sql
@@ -1,0 +1,1 @@
+ALTER TABLE projects DROP COLUMN updated_time;
--- a/server/src/optima/dbmodels.py
+++ b/server/src/optima/dbmodels.py
@@ -46,7 +46,9 @@ class ProjectDb(db.Model):
                                 uselist=False)
     project_data = db.relationship('ProjectDataDb', backref='projects',
                                 uselist=False)
-    creation_time = db.Column(db.DateTime(timezone=True), server_default=text('now()'))
+    creation_time = db.Column(db.DateTime(timezone=True), default=db.func.now())
+    updated_time = db.Column(db.DateTime(timezone=True),
+                             default=db.func.now(), onupdate=db.func.now())
 
     def __init__(self, name, user_id, datastart, dataend, \
         programs, populations, model = None, creation_time = None):
--- a/server/src/optima/dbmodels.py
+++ b/server/src/optima/dbmodels.py
@@ -46,9 +46,8 @@ class ProjectDb(db.Model):
                                 uselist=False)
     project_data = db.relationship('ProjectDataDb', backref='projects',
                                 uselist=False)
-    creation_time = db.Column(db.DateTime(timezone=True), default=db.func.now())
-    updated_time = db.Column(db.DateTime(timezone=True),
-                             default=db.func.now(), onupdate=db.func.now())
+    creation_time = db.Column(db.DateTime(timezone=True), server_default=text('now()'))
+    updated_time = db.Column(db.DateTime(timezone=True), onupdate=db.func.now())
 
     def __init__(self, name, user_id, datastart, dataend, \
         programs, populations, model = None, creation_time = None):
--- a/client/source/js/modules/home/home-ctrl.js
+++ b/client/source/js/modules/home/home-ctrl.js
@@ -9,6 +9,7 @@ define(['./module', 'underscore'], function (module) {
 
       if ($scope.project) {
         $scope.project.creation_time = Date.parse($scope.project.creation_time);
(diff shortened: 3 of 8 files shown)


dragansah:
https://trello.com/c/dDlUCAgD/599-redo-project-metadata-table
and
https://trello.com/c/pFlAnnIv/592-replace-data-uploaded-on-with-project-last-updated

server/db/versions/010_add_updated_time_field_postgresql_downgrade.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE projects DROP COLUMN updated_time;
```
Ashalynd: This is not very nice: that means you'll remove the updated time for all existing projects. Also, I don't get why this change is applied at all, what was the reason for that?

server/src/optima/dbmodels.py
```diff
@@ -46,7 +46,9 @@ class ProjectDb(db.Model):
                                 uselist=False)
     project_data = db.relationship('ProjectDataDb', backref='projects',
                                 uselist=False)
-    creation_time = db.Column(db.DateTime(timezone=True), server_default=text('now()'))
+    creation_time = db.Column(db.DateTime(timezone=True), default=db.func.now())
```
Ashalynd: We had updated_time = creation_time initially, which was assumed to be confusing.

Ashalynd:
I suggest not to change the existing structure of the tables, because it does not seem to be needed for this, and can introduce problems for the existing projects.

server/db/versions/010_add_updated_time_field_postgresql_downgrade.sql
Ashalynd: Oops, didn't realize that was the new column.

server/src/optima/dbmodels.py
Ashalynd: This change for creation_time does not make sense, because already existing column in DB is not modified anyway. For updated_time, we don't need the default (because when the project is created, it has not been updated yet).

tdelev: The change for `creation_time` it's for consistency reasons, the effect of both is same and I think the later it's more robust because allows defaults on INSERT and UPDATE. Maybe only we could add to the migration script drop defaults
```sql
ALTER COLUMN creation_time DROP DEFAULT;
```
The defaults for `updated_time` it's not problematic, we could remove, but still it's no source of confusion since same `creation_time` and `updated_time` means no update.

Ashalynd: Let's not change the existing stuff that is proven to work, also I'd rather not drop DEFAULT value for data creation column just for the sake of ORM :-}

dragansah: The change on `creation_time` is for consistency reasons, and also I think using `default` instead of `server_default` is more robust because it allows defaults on INSERT and UPDATE. Only think we could add is one line in the migration script, to align the database:
```sql
ALTER COLUMN creation_time DROP DEFAULT;
```
For `updated_time` the default is not problematic, and can be removed, but still it's not source of confusion because same `creation_time` and `updated_time` means no update.

dragansah: The change of the database is required to complete the task, it is the task. And it will not break any existing stuff.

Ashalynd: I am ok with introducing new column, but there are no real reasons to change the creation_time (creation_time is only set once for a new record and should never be updated), so I suggest leaving that column the way it was.

Apart from that I'd like to get confirmation from Cliff that he really wants to log every update to the _project_  vs every update to the _model_ :-}

cliffckerr: I don't know what distinction you're making between the "project" and the "model"...any time anything is saved the "Last modified" date should be updated.

Ashalynd: Project is updated when you e.g. upload a new json.

Model is updated when you run a new calculation and save the result.

Ashalynd: Regarding the difference between default and server_default (which we consistently have in many places already :) )  server_default means you don't provide the value to the column and DB fills the default there; default means you specify which default value to provide in 'INSERT' command. I prefer the first approach because it guarantees that there will always be a value if something is inserted.

cliffckerr: If you save the result then you modify the json :) I still don't think there's a difference, and "Last updated" should reflect any change in either the "model" or the "project".

dragansah:
@Ashalynd implemented your remarks by returning `server_default` on `creation_date` field and removed `default` on `updated_time`, @cliffckerr here is a screenshot of projects list 
no_mod

Ashalynd:
Thanks!

PR #413, opened 2015-04-05
Feature/refactoring

nikgraf:
* Aligned the way we deal with a model that is not ready due it's missing data
* Fixed revering the model in cost coverage
* Fixed saving the model in cost coverage
* Fixed possible reference errors by moving everything into a state object in cost coverage & optimization(Angular best practice)
* Removed legacy validation in cost coverage
* Added tests for cost coverage service


Ashalynd:
The functionality works (verified that).

PR #415, opened 2015-04-10
refactor(project): deprecate legacy unused attributes

nikgraf:
https://trello.com/c/7dehOBvq/602-remove-legacy-can-scenarios-can-calibrate

Since the model is generated directly with the upload we don't need to check if model parameters exists.

### DIFF
--- a/client/source/js/modules/model/calibration-ctrl.js
+++ b/client/source/js/modules/model/calibration-ctrl.js
@@ -26,7 +26,7 @@ define(['./module', 'angular', 'underscore'], function (module, angular, _) {
 
     var initialize = function() {
       $scope.projectInfo = info;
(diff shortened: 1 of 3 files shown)


client/source/js/modules/model/calibration-ctrl.js
```diff
@@ -26,7 +26,7 @@ define(['./module', 'angular', 'underscore'], function (module, angular, _) {
 
     var initialize = function() {
```
Ashalynd: so it's false by default? is this variable used?

nikgraf: yes, it waits until the data is back from the sever before allowing the user to select automatic or manual calibration. I thought about removing it entirely, because nobody will do it. I still kept it as it felt the right thing to do.

PR #420, opened 2015-04-19
fix yscale on cost-outcome graphs

robynstuart:


Ashalynd:
@robynstuart the issue seems to be fixed now :)

PR #423, opened 2015-04-24
Feature/567 investigate mpld3

nikgraf:
https://trello.com/c/m5CIoJvJ/567-investigate-mpld3

- The chart bar is now extended by reset. That's why the name changed to chart-toolbar.
- Added the latest mpld3 version including a bug fixed which already got merged.
- Replaced custom d3 charts with generated mpld3 charts.


server/src/sim/utils.py
```diff
@@ -223,13 +225,25 @@ def dumpfile(variable):
         inds = argsort(printnames)
     else:
         inds = argsort(printbytes)
-    
+
     for v in inds:
         print('Variable %s is %s' % (printnames[v], printsizes[v]))
-    
+
```
Ashalynd: Is this one using globals?

Ashalynd: (fine until checkmem is not being called from any web function, I guess)

Ashalynd:
Verified that it works on my machine. BE should be fine. Please @dmitryevseev and/or @sebastianconcept review the FE

dmitryevseev:
I'm taking a look

server/src/sim/utils.py
cliffckerr: checkmem is just a helper function, not even used on the BE, but useful for e.g. looking at where the JSON files blow out in size

PR #424, opened 2015-04-28
Feature/574 add hotkeys to dialogs

sebastianconcept:
Pairing with @jpuri on this card
https://trello.com/c/tFpuBZ7d/574-add-hotkeys-to-dialogs

### DIFF
--- a/client/source/js/modules/ui/modal/modal-service.js
+++ b/client/source/js/modules/ui/modal/modal-service.js
@@ -21,15 +21,30 @@ define([
          * executes the right callback depending on the user's choice.
          */
         confirm: function (onAccepted, onRejected, acceptButton, rejectButton, message, title) {
-          $modal.open({
+
+          var onModalKeyDown = function (anEvent) {
+                debugger
+                if(anEvent.keyCode == 13) { return modalInstance.close(true) } // ENTER
+                if(anEvent.keyCode == 78) { return } // N
+                if(anEvent.keyCode == 89) { return } // Y
+          };
+
+          var modalInstance = $modal.open({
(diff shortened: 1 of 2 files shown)


client/source/js/modules/ui/modal/modal-service.js
```diff
@@ -21,15 +21,24 @@ define([
          * executes the right callback depending on the user's choice.
          */
         confirm: function (onAccepted, onRejected, acceptButton, rejectButton, message, title) {
-          $modal.open({
+
+          var onModalKeyDown = function (anEvent) {
```
nikgraf: @jpuri @sebastianconcept Why `anEvent` instead of just `event`?

nikgraf: Just found this: http://stackoverflow.com/a/1510106/837709

sebastianconcept: The convention of using `a` or `an` or `someBlah` is explicitely referring to an instance(s). When used consistently, it makes argument naming less ambiguous.

nikgraf:
Works like a charm! thx @sebastianconcept @jpuri

client/source/js/modules/ui/modal/modal-service.js
sebastianconcept: And I forgot to mention that, that naming together with good method naming, makes it closer to spoken english:
`this.doSomethingWith(anObject)`
`this.convertUsing(aConverter)`
`moneyController.includes(somePayments)`
`carModel.removeTires(someTires)`
etc

PR #426, opened 2015-04-29
increasing line spacing on optimization analysis page

jpuri:
The PR has fix for:
https://trello.com/c/MYWa4xlq/593-inadequate-line-spacing


nikgraf:
@jpuri nice!

small note: we usually use dashes for classes instead of underscores

PR #427, opened 2015-04-30
Feature/621 remove the 5th option from cost coverage chart

jpuri:
The PR has fix for this issue:

https://trello.com/c/p2FsooQi/621-remove-the-5th-option-from-cost-coverage-chart

The param is still mentioned in 004_fix_ccparams.py, but I think that old insignificant file.


nikgraf:
looks good, thx for the state fix

PR #428, opened 2015-05-01
plot current param vals in red

robynstuart:


nikgraf:
@robynstuart works like a charm

PR #431, opened 2015-05-04
Increasing width of text boxes on optimization analyses page

jpuri:
The PR has fix for:
https://trello.com/c/qtqI80p7/624-example-value-on-optimization-analysis-define-constrains-page-is-not-visible-in-1920x1080


jpuri:
I was also confused about using large size for that text box. Though its works well on MAC but looks like with resolution that Natalia uses (1920x1080), probably bigger text boxes look better.

PR #434, opened 2015-05-06
632-it-s-not-possible-to-zoom-move-and-reset-after-reverting-model

jpuri:
The PR has fix for this issue:
https://trello.com/c/5FOIHkrs/632-it-s-not-possible-to-zoom-move-and-reset-after-reverting-model


nikgraf:
@jpuri I thought we took care of that by making sure we generate a new ID all the time. I noticed I missused ng-init & for the cost-coverage chart it failed to work.

I updated PR so it works now. I removed the splicing as I would like to avoid changing library internal data structures. There is a chance of messing up other stuff by doing so.

PR #435, opened 2015-05-06
option minimize money not working on optimization page

jpuri:
The PR has fix for this big:
https://trello.com/c/aEo6k9YZ/626-cost-assumptions-tabs-are-not-active


nikgraf:
good one @jpuri 

we missed that during the last refactoring

PR #440, opened 2015-05-08
initialize targetpopmodel before condition

mkhurramali:
https://trello.com/c/EBlSa8fn/627-error-messages-on-be-mac-only


nikgraf:
@robynstuart is this change fine?

PR #441, opened 2015-05-10
Feature/remove mpld3 from sim

nikgraf:
https://trello.com/c/My1IieTK/638-remove-mpld3-from-src-sim

As discussed with @Ashalynd we are introducing a plotting.py in server. In sim we will generate & return matplotlib figures to be plotted. Hint: for the sim BE we can't close the figures as they closing would mean you won't see the plot in the BE.

For the FE we close the figures in plotting.py after retrieved the necessary data.

@Ashalynd & @robynstuart can you review these changes?
@cliffckerr With this PR we would remove mpld3 from sim again.
@robynstuart I was running `python run_makeccocs.py -p -w` to see if the BE works & charts get plotted. Is this correct?


robynstuart:
@nikgraf i haven't used run_makeccocs.py for awhile, but i understand it's used in testing.

PR #452, opened 2015-06-06
Reverting pr 427

jpuri:
Hey @robynstuart ,

The PR has code changes related to setting upper limit on x-axis reverted.

One more thing I noticed while redoing the changes is that on page 
"Define cost-coverage-outcome assumptions"

The optional parameters (3,4,5,6,7) are passed only when params 1,2 have some value, but I am not sure if its a desired behavior or a bug.


robynstuart:
thanks @jpuri, looks good!

PR #454, opened 2015-06-07
Oop cyclic

RomeshA:
@davidkedz 

So this proposal incorporates a cyclic reference in the Simbox and Sim classes. It's not that weird - for example, a doubly-linked list structure (one of the most basic data types they teach in CS) has objects containing references to both their parent and child. The problems this solves are

- You no longer have to include multiple arguments in a number of places, because the properties can be obtained from the region directly
- You can no longer introduce bugs by mixing bits and pieces of different regions in said function calls
- By interacting more directly with the Simboxes and Sims, there is a reduced need for wrapper methods in Region(). This is a big issue, because the wrapper methods make the region a (mentally) heavier class, and because region.py will be changed much more frequently than is necessary if all clean syntax requires helper methods
- The simulation only makes sense in terms of its parent region - this is enforced by having only one region reference within the Sim. With the current structure, you can pass in data from one region, metadata from another region etc. and thus put the Sim into an invalid state. This would most likely occur inadvertently/unintentionally

The biggest questions about storing the references to the parent region are

- How do we make sure that data isn't duplicated, both in memory and on disk
- How do we make sure that garbage collection still works properly

**Making sure data isn't duplicated**
The Simboxes and Sims use the weakref() module (Python builtin) to store what is essentially a pointer to the region. This pointer behaves in a similar way to pointers in C/C++ in that the data for the region is not duplicated. The getregion() method resolves the pointer, and it is called within the class methods as required. Calling getregion() also means that the Simboxes and Sims will always look up the current state of the region they are contained. Saving and loading files is currently accomplished by the todict() and fromdict() methods. The todict() method records the region UUID rather than the reference to the region. A Sim and Simbox can only be used in the context of a region. The fromdict() method therefore also requires a region argument. The region that is passed in needs to correspond to the original parent region, so fromdict() checks the UUID to make sure they match. Currently, this functionality is all wrapped in the region save/load methods, so it is transparent to the user. However, in future this functionality will make it possible to save and load Sims and Simboxes to be shared without sending the region, as long as both users already have the same parent Region object. 

**Garbage collection**
The weakref() class is designed to work properly with garbage collection. Specifically, it does not count as a full reference to the object for garbage collection purposes, and thus it will not stop the parent from being destroyed. This means that the failure mode is the getregion() method returning a NoneType object. We can check for this in future and print a helpful message. For most usages, everything will work fine. One usage pattern that is not supported would be

    def simfactory():
        r = Region()
        s = Sim('test,r)
        return s

But this is fine, because a Sim only makes sense in the context of the region. So the weakref system will enforce our intended usage

---

So...I wasn't initially sold on having cyclic references either because a number of potential problems came to mind. I made several mistakes already in passing around the region data manually, and am increasingly thinking that the extensive wrapper methods are adding unnecessarily complexity. I think that the implementation of cyclic references here ended up being much easier than we initially envisaged and can avoid the problems we previously discussed. Thoughts? (let's not decide immediately)

### DIFF
--- a/server/src/sim/classes/portfolio.py
+++ b/server/src/sim/classes/portfolio.py
@@ -14,8 +14,6 @@ import os
 from numpy import arange
 
 from region import Region
-import defaults
-
 
 class Portfolio:
     def __init__(self, portfolioname):
@@ -31,7 +29,6 @@ class Portfolio:
            
         Version: 2015may28 by davidkedz
         """
-        #from time import time
         
         print('\nPortfolio %s has been activated.' % self.portfolioname)
         print('\nThe script that created this portfolio is in...')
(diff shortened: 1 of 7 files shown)


RomeshA:
The concept seems to be working alright - shall we merge this now?

cliffckerr:
Umm...when I tried to run Optima via the frontend on this branch it died horribly :) Are you sure it's ready?

cliffckerr:
Also, since these features aren't ready to be used by end users, is there an advantage to merging right now? Anything committed to develop goes live immediately to all the frontend users, and non-frontend users can just switch to the appropriate Git branch...

RomeshA:
This pull request is to merge oop_cyclic into oop_experiment, not into develop, so this is mainly for @davidkedz - apologies for the confusion

PR #607, opened 2016-01-27
Implement ost

robynstuart:
@cliffckerr please review.

• fixed bug in constants in spreadsheet
• implemented OST & added a scenario to check effect (which is small, but not inconceivably so)


optima/makespreadsheet.py
```diff
@@ -681,9 +681,9 @@ def generate_const(self):
             'Unsuppressive ART',
             'Suppressive ART',
             'Probability of viral suppression on ART'],
-            [0.95, 0.58, 0.54, 0., 2.65, 0.9, 0.73, 0.5, 0.92, 0.9], 
-            [0.8, 0.47, 0.33, 0., 1.35, 0.82, 0.65, 0.3, 0.8, 0.8 ],
-            [0.98, 0.67, 0.68, 0.68, 5.19, 0.93, 0.8, 0.8, 0.95, 0.95],
+            [0.95, 0.58, 0., 2.65, 0.54, 0.9, 0.73, 0.5, 0.92, 0.9],
```
cliffckerr: these are still not right...

optima/model.py
```diff
@@ -90,26 +90,43 @@ def model(simpars=None, settings=None, verbose=2, safetymargin=0.8, benchmark=Fa
...
+    circum   = simpars['circum']    # Prevalence of circumcision (P)
+    stiprev  = simpars['stiprev']   # Prevalence of STIs (P)
+    prep     = simpars['prep']      # Prevalence of PrEP (P)
+
+    # Uptake of OST
+    numost = simpars['numost']                  # Number of people on OST (N)
+    if any(injects):
+        numpwid = popsize[injects].sum(axis=0)  # Total number of PWID
+        try: ostprev = numost/numpwid           # Proportion of PWID on OST (P)
+        except: raise Exception('Cannot divide by the number of PWID')
```
cliffckerr: please use OptimaException instead of Exception from now on (requested by SSQ)

optima/parameters.py
```diff
@@ -531,7 +531,7 @@ def interp(self, tvec, smoothness=20):
             output = zeros((npops,len(tvec)))
             for pop,key in enumerate(keys): # Loop over each population, always returning an [npops x npts] array
                 output[pop,:] = self.m * smoothinterp(tvec, self.t[pop], self.y[pop], smoothness=smoothness) # Use interpolation
-        if npops==1: return output[0,:]
+        if npops==1 and self.by!='pship': return output[0,:]
```
cliffckerr: so i assume this is to fix it in cases where there's a single partnership, but it'll still fail for a single population, so maybe `if npops==1 and self.by!='pship'` should be replaced by `if self.by=='tot'`?

tests/testgui.py
```diff
@@ -54,7 +54,7 @@ def done(t=0):
     print('Running plot test...')
     from optima import Project, plotresults
     
-    P = Project(spreadsheet='test7pops.xlsx')
+    P = Project(spreadsheet='testpwid.xlsx')
```
cliffckerr: i'd personally prefer to keep these more generic...

optima/makespreadsheet.py
robynstuart: no? what should they be?

optima/parameters.py
robynstuart: ok

tests/testgui.py
robynstuart: what if we just have a few standard ones (PWID-driven, SW-driven, GP)?

optima/makespreadsheet.py
cliffckerr: will have to check but upper limit for diagnosis behavior change of 5.19 looks wrong to me, if i'm reading that correctly

tests/testgui.py
cliffckerr: yes, sounds good

optima/makespreadsheet.py
robynstuart: true. that's for STIs.

PR #609, opened 2016-01-27
Coverage plots

robynstuart:
@cliffckerr please review. Just adding a coverage plot and fixing the floats issue with budget.


optima/plotting.py
```diff
@@ -429,56 +429,63 @@ def plotimprovement(results=None, figsize=(14,10), lw=2, titlesize=14, labelsize
...
-    budgetyearstoplot = [budgetyears for budgetyears in results.budgetyears.values() if budgetyears]
-    proglabels = budgetstoplot[0].keys() 
-    alloclabels = [key for k,key in enumerate(results.budget.keys()) if results.budget.values()[k]] # WARNING, STUPENDOUSLY UGLY
-    nprogs = len(proglabels)
-    nallocs = len(alloclabels)
+    for plotkey in which:
+        toplot = budgetstoplot if plotkey=='budget' else coveragetoplot
+        proglabels = toplot[0].keys() 
+
+        alloclabels = [key for k,key in enumerate(multires.__dict__[plotkey].keys()) if multires.__dict__[plotkey].values()[k]]
```
cliffckerr: i generally prefer
`getattr(multires, plotkey)`
to
`multires.__dict__[plotkey]`
since shorter and slightly more transparent...

optima/results.py
```diff
@@ -102,7 +102,7 @@ def __repr__(self):
 
 class Resultset(object):
     ''' Structure to hold results '''
-    def __init__(self, raw=None, name=None, simpars=None, project=None, settings=None, data=None, parset=None, progset=None, budget=None, budgetyears=None, domake=True):
+    def __init__(self, name=None, raw=None, simpars=None, project=None, settings=None, data=None, parset=None, progset=None, budget=None, coverage=None, budgetyears=None, domake=True):
```
cliffckerr: i know it's inconsistent, but i had it this way because usually you _don't_ want to specify a name for a resultset, and the other syntax meant that you could do Resultset(raw), which is all you actually need -- so here i agree it would be more transparent if raw were _not_ a kwarg, except then you couldn't do the syntax of Resultset(name='foo', raw=raw) as you would with other objects -- so i'd prefer to leave it as it was

robynstuart: I did not change this. But I did have a few merge conflicts, so perhaps this was one of them

cliffckerr:
merged into minimize-money

PR #613, opened 2016-01-28
Minimize money

cliffckerr:
This branch has...a lot of changes -- it implements minmoney() and changes how many, many other things work. @robynstuart and @davidkedz and @nanuxbe / @Ashalynd please review. 

NB: this branch incorporates geogui-with-plotfixes and coverage-plots branches as well, since those branches had too many conflicts with this branch to merge into develop separately.


optima/geospatial.py
```diff
@@ -0,0 +1,335 @@
...
+    def makesheet():
+        ''' Create a geospatial spreadsheet template based on a project file '''
+        
+        ## 1. Load a project file
+        project = _loadproj()
+            
+        ## 2. Get destination filename
+        spreadsheetpath = QtGui.QFileDialog.getSaveFileName(caption='Save geospatial spreadsheet file', filter='*.xlsx')
+        
+        ## 3. Extract data needed from project (population names, program names...)
```
robynstuart: is this not ready yet? if so, please add a todo in the code so it's clear whether there are pieces missing here

optima/geospatial.py
```diff
@@ -0,0 +1,335 @@
...
+        destination = QtGui.QFileDialog.getExistingDirectory(caption='Choose output folder')
+        
+        ## 4. Read the spreadsheet
+        # ...
+        
+        ## 5. Calibrate each project file according to the data entered for it in the spreadsheet
+        # ...
+        
+        ## 6. Save each project file into the directory
+        # ...
```
robynstuart: see above

optima/geospatial.py
```diff
@@ -0,0 +1,335 @@
...
+        
+        return None
+
+
+    def create():
+        ''' Create a portfolio by selecting a list of projects; silently skip files that fail '''
```
robynstuart: why would we want to silently skip files that fail?

tests/benchmark.py
```diff
@@ -0,0 +1,40 @@
+"""
+BENCHMARK
+
```
robynstuart: I'd prefer that this file was called something like benchmarkmodel or timemodel so that it's clear from the name what it is meant for. There could be other things that we'd want to benchmark.

optima/project.py
```diff
@@ -382,9 +383,69 @@ def runbudget(self, budget=None, budgetyears=None, progsetname=None, parsetname=
...
+        self.addresult(result=multires)
+        self.modified = today()
+        return None
+        
+        
+    #######################################################################################################
+    ## Methods to handle specialised tasks (i.e. for geospatial analysis)
+    #######################################################################################################
+        
+    def genBOC(self, budgetlist=[10000,100000,1000000,10000000], name=None, parsetname=None, progsetname=None, inds=0, objectives=None, constraints=None, maxiters=1000, maxtime=None, verbose=5, stoppingfunc=None, method='asd'):
```
robynstuart: Shouldn't really pass in a list like this, I think. Also, from what I can see, this is a list of the total budgets that the individual projects will be allocated (in the loop below), is that right? If so, wouldn't we (1) want higher numbers (e.g. Zambia's budget is ~$500m) and/or (2) want to use a scaling factor of the actual total budget rather than setting an amount?

tests/benchmark.py
cliffckerr: done in the fix-geogui PR

optima/geospatial.py
cliffckerr: so if the user is selecting 30 files and accidentally gets a wrong one, it doesn't kill everything; the user can add missing ones in later...

optima/geospatial.py
cliffckerr: i usually use 'todo' for 'get to eventually' but can add, good idea

optima/geospatial.py
```diff
@@ -0,0 +1,339 @@
+"""
+GEOSPATIAL
```
robynstuart: NB I'm ignoring most of this file as I understand it's been superseded by the next PR.

optima/model.py
```diff
@@ -110,11 +110,26 @@ def model(simpars=None, settings=None, verbose=2, safetymargin=0.8, benchmark=Fa
...
     if any(injects):
         numpwid = popsize[injects,:].sum(axis=0)  # Total number of PWID
-        try: ostprev = numost/numpwid           # Proportion of PWID on OST (P)
-        except: raise OptimaException('Cannot divide by the number of PWID')
-    else:
-        if sum(numost): raise OptimaException('You have entered non-zero value for the number of PWID on OST, but you have not specified any populations who inject')
-        else: ostprev = 0.
+        try: 
+            ostprev = numost/numpwid # Proportion of PWID on OST (P)
+            ostprev = minimum(ostprev, 1.0) # Don't let more than 100% of PWID be on OST :)
```
robynstuart: @cliffckerr this is ok for now but should not really be here. We could add a check like this in loadspreadsheet and notify users that way. Or we could add a check for it in parameters.py - upper limit of numost should be numpwid

optima/programs.py
```diff
@@ -7,33 +7,13 @@
...
-    else:
-        errormsg = 'Could not convert budget vector into budget: incompatible lengths (%i vs. %i)' % (len(budgetvec), len(budget))
-        raise OptimaException(errormsg)
-    
-    return budget
 
+# WARNING, this should not be hard-coded!!! Available from
+# [par.coverage for par in P.parsets[0].pars[0].values() if hasattr(par,'coverage')]
+# ...though would be nice to have an easier way!
+coveragepars=['numtx','numpmtct','numost','numcircum']
```
robynstuart: @cliffckerr agreed. At some stage, it'd be nice it Programs stored actual parameter objects, not just strings that are used to access them. We can do this at a later stage.

optima/results.py
```diff
@@ -100,9 +102,49 @@ def __repr__(self):
 
 
 
+class BOC(object):
+    ''' Structure to hold a budget and outcome array for geospatial analysis'''
+    def __init__(self, name='unspecified', x=None, y=None, objectives=None):
```
robynstuart: @davidkedz of course it's not urgent, but it would be nice to rename x and y something like budget and outcome, or totalspends and epioutcomes, or anything else that would give a clue about what they should be.

robynstuart:
@cliffckerr @davidkedz am happy for this to be merged, once @nanuxbe or @Ashalynd has confirmed that it won't break anything.

nanuxbe:
@robynstuart I get this error when going to the calibration screen
```
Traceback (most recent call last):
  File "/home/emma/dev/Optima/server/webapp/utils.py", line 63, in _report_exception
    return api_call(*args, **kwargs)
  File "/home/emma/dev/Optima/server/p-env/local/lib/python2.7/site-packages/flask_restful/__init__.py", line 679, in wrapper
    resp = f(*args, **kwargs)
  File "/home/emma/dev/Optima/server/webapp/resources/project_parsets.py", line 271, in get
    selectors = self._selectors_from_result(result, which)
  File "/home/emma/dev/Optima/server/webapp/resources/project_parsets.py", line 223, in _selectors_from_result
    graph_selectors = op.getplotselections(result)
  File "/home/emma/dev/Optima/optima/plotting.py", line 58, in getplotselections
    if hasattr(results, budcov) and len(getattr(results, budcov))>0:
TypeError: object of type 'NoneType' has no len()
```

PR #617, opened 2016-01-29
Fix geogui

cliffckerr:
Added significant additional functionality to the geospatial GUI, including project info, ability to add or remove projects, etc. Many bugfixes too. Things that don't work: top 2 buttons and export results. All else should work, including running geo analysis. Please review @robynstuart or @davidkedz .

@davidkedz I still really don't like the splines, I think the first derivatives are awful (see notes.optimamodel.com, today's date), and would prefer to replace with something smoother -- I tried smoothinterp (and added it as a method to pchip, lol) but that failed because you ask it to calculate instantaneous derivatives, and I didn't have time to fix that for the smoothinterp method.


robynstuart:
@cliffckerr thanks, looks great!

PR #621, opened 2016-01-29
Births

robynstuart:
@cliffckerr please review.

Implements births, MTCT, PMTCT, age transitions, risk transitions. 

Still to do: circs, check that everything makes sense, check that PMTCT has appropriate effects.


optima/model.py
```diff
@@ -105,6 +107,8 @@ def model(simpars=None, settings=None, verbose=2, safetymargin=0.8, benchmark=Fa
     circum   = simpars['circum']    # Prevalence of circumcision (P)
     stiprev  = simpars['stiprev']   # Prevalence of STIs (P)
     prep     = simpars['prep']      # Prevalence of PrEP (P)
+    numpmtct = simpars['numpmtct']  # Number (or proportion?) of people receiving PMTCT (P/N)
+    usepmtctprop=True if all(numpmtct<1) else False
```
cliffckerr: this reminds me of what we had in optima 1.0 which...didn't work :(

optima/parameters.py
```diff
@@ -344,7 +360,7 @@ def makepars(data, label=None, verbose=2):
     # Set up keys
     totkey = ['tot'] # Define a key for when not separated by population
     popkeys = data['pops']['short'] # Convert to a normal string and to lower case...maybe not necessary
-    fpopkeys = [popkeys[i] for i in range(len(popkeys)) if pars['female'][i]]
+    fpopkeys = [popkey for popno,popkey in enumerate(popkeys) if data['pops']['female'][popno]]
```
cliffckerr: sorry, what's the reason for this change?

optima/parameters.py
```diff
@@ -386,26 +402,43 @@ def makepars(data, label=None, verbose=2):
     ## Tidy up -- things that can't be converted automatically
     ###############################################################################    
     
-    # Births
+    # Births rates. This parameter is coupled with the birth matrix defined below
     for key in list(set(popkeys)-set(fpopkeys)): # Births are only female: add zeros
         pars['birth'].y[key] = array([0])
         pars['birth'].t[key] = array([0])
+    pars['birth'].y = pars['birth'].y.sort(popkeys) # Sort them so they have the same order as everything else
+    pars['birth'].t = pars['birth'].t.sort(popkeys)
```
cliffckerr: ah, good point -- this probably needs to be done for circumcision etc. as well

optima/parameters.py
```diff
@@ -462,7 +496,8 @@ def makesimpars(pars, inds=None, keys=None, start=2000, end=2030, dt=0.2, tvec=N
     
     # Copy default keys by default
     for key in generalkeys: simpars[key] = dcp(pars[key])
-    
+    for key in staticmatrixkeys: simpars[key] = array(pars[key])
```
cliffckerr: should probably dcp, no?

tests/testmakespreadsheet.py
```diff
@@ -9,7 +9,7 @@
 ## Define tests to run here!!!
 tests = [
 'makespreadsheet',
-#'unittests',
+'unittests',
```
cliffckerr: the problem with the unittests is that they kill the console (at least for me...?)

optima/model.py
robynstuart: no, what didn't work in v1 was that we checked the entries separaterly on each time loop, like
`if numpmtct[t]>1` for each t.
The broke because it could have been 0 at one time step but a number the next. I think it'd valid to do it this way, however, where you check the whole array at once.

optima/parameters.py
robynstuart: I wanted to see if it was the same thing :) actually I did change it back, but i see that I forgot to push.

tests/testmakespreadsheet.py
robynstuart: not for me. all the GUIs do however!!!

optima/model.py
cliffckerr: it still feels very fragile and hacky to me; i'd prefer to have separate parameters, but that has its pitfalls too. i don't have a better suggestion so i'm ok to leave it as is for now.

PR #624, opened 2016-01-30
Add cost functions to a program

hawkowl:
see https://trello.com/c/bbBAT464/698-6-6-1-add-cost-functions-to-a-program, cc @Ashalynd


Ashalynd:
Update API would be done in a different way (on the same card)

PR #635, opened 2016-02-01
Feature/743 implement automatic fitting

nanuxbe:
- Basic APi only
- Celery config + dependencies
- Celery task
- Deploy script


server/README.md
```diff
@@ -3,88 +3,28 @@ Introduction
...
-Project API
-=====
+`sudo apt-get install redis-server`
 
-These APIs allow front-end to work with projects.
+On MacOS X:
+---
 
-* `/api/project/info`
+Follow [these instructions](http://jasdeep.ca/2012/05/installing-redis-on-mac-os-x/)
```
Ashalynd: brew install redis

Ashalynd:
It looks great! A couple of remarks / questions:

1) the piece about actually running celery locally is missing - would be good to add it to Readme. E.g. what to do with that example config ?

2) could it be possible to have celery.sh also in server folder, are there any reasons for a different setup?

server/README.md
cliffckerr: please also update the main README.md for anything essential -- the subreadmes should be additional info, but the README in the root dir should be sufficient to get up and running

PR #636, opened 2016-02-01
Geospatial bugfixes

cliffckerr:
fixes several bugs in portfolio.py


cliffckerr:
@robynstuart please review, bugs fixed

robynstuart:
Export results sends results to excel with numbers stored as text. Can this be fixed?

cliffckerr:
yes -- so it was a number represented as a string being stored as a number which actually stored it as a string. now it's a number represented as a number being stored as a number, which works.

robynstuart:
Would also be great to have "Outcome" disaggregated into "Infections averted" and "Deaths averted" or "Cumulative infections" vs "Cumulative deaths" or any other more meaningful thing

cliffckerr:
agreed

robynstuart:
thanks @cliffckerr @davidkedz !!

PR #637, opened 2016-02-01
Econ data: validation to check if user uploaded regular datasheet

guptaarun:
- Cannot load economic datasheet without first uploading regular datasheet
https://trello.com/c/fH0ptDrp/808-cannot-load-economic-datasheet-without-first-uploading-regular-datasheet


robynstuart:
thanks @guptaarun looks good :)

PR #640, opened 2016-02-02
Fix condom parameters

robynstuart:
@cliffckerr or @davidkedz please review - fixes the duplication of partnership parameters for condoms. Note that acts are still duplicated.


cliffckerr:
it seems to work so i'm merging it, but note that this has NOT fixed the frontend -- the pairs are still visible. but a bug card can be made for that.

robynstuart:
@cliffckerr it looks to me like it HAS fixed the FE, as well it should because the FE displays the keys of parameter.y. 
screen shot 2016-02-02 at 5 46 29 am

PR #642, opened 2016-02-02
Show error message if deleting the default parameter set

boscoh:


cliffckerr:
both features work great, thanks @boscoh !

PR #644, opened 2016-02-02
Circumcision bugfixes

cliffckerr:
major fixes to scenarios such that targeted populations are chosen instead of random populations :) also introduces a new gui function, plotpars() that will compare any number of objects like P.parsets[0].pars[0]


cliffckerr:
subsumed by new PR

PR #646, opened 2016-02-02
Diagnosing bumpiness

cliffckerr:
Fixed the major bugs with circumcision. WARNING: aging does NOT work since the circ/uncirc balance is hard-reset on each timestep in proportion with `circum[pop,t]`.Everything else should work. @robynstuart please review.


optima/model.py
```diff
@@ -7,18 +7,19 @@ def model(simpars=None, settings=None, verbose=None, benchmark=False, die=True):
...
     if benchmark: starttime = tic()
     
+    
     ###############################################################################
     ## Setup
     ###############################################################################
 
     # Hard-coded parameters that hopefully don't matter too much
-    eps = 1e-3 # Define another small number to avoid divide-by-zero errors
+    eps = 1.0101e-3 # Define another small number to avoid divide-by-zero errors
```
robynstuart: was there a problem with the previous eps?

cliffckerr: i kept getting these incorrect results of 1e-3 and i thought it might be an eps, but it wasn't -- anyway, this has been superseded anyway, eps is now set by settings.eps

optima/gui.py
```diff
@@ -553,4 +553,134 @@ def plotpeople(project=None, people=None, start=2, end=None, pops=None, animate=
...
+                    except: print('??????')
+                    try: 
+                        if not(hasattr(this[3],'__len__') and len(this[3])==0): ax.scatter(this[2],this[3])
+                    except Exception: pass # print('Problem with "%s": "%s"' % (this[0], E.message))
+                    if pd==len(allplotdata)-1: # Do this for the last plot only
+                        ax.set_ylim((0,1.1*ax.get_ylim()[1]))
+                        ax.set_xlim((tvec[0],tvec[-1]))
+                
+    update()
+    plotparsbackbut.on_clicked(updateb)
```
robynstuart: not functional...

PR #651, opened 2016-02-02
Persisting celery

nanuxbe:


server/webapp/dbmodels.py
```diff
@@ -360,13 +362,16 @@ class WorkLogDb(db.Model):  # pylint: disable=R0903
...
     project_id = db.Column(UUID(True), db.ForeignKey('projects.id'))
+    parset_id = db.Column(UUID(True))
+    result_id = db.Column(UUID(True), default=None)
     start_time = db.Column(db.DateTime(timezone=True), server_default=text('now()'))
     stop_time = db.Column(db.DateTime(timezone=True), default=None)
     status = db.Column(work_status, default='started')
     error = db.Column(db.Text, default=None)
 
-    def __init__(self, project_id, work_type=None):
+    def __init__(self, project_id, parset_id = parset_id, work_type=None):
```
Ashalynd: The rest should be filled in automatically.

nanuxbe: yes, but parset_id=parset_id?  there is no global var named parset_id

Ashalynd: There is such property in that table.

Ashalynd: Ah right, I see what you mean. Indeed, you are right!

PR #654, opened 2016-02-03
Cleaned up imports

cliffckerr:
ok, i should've read up on relative imports a bit more before making that horrifying optima.py :) this branch should have absolutely no impact on anything, just putting imports into __init__.py where they belong. @Ashalynd and/or @robynstuart please review.

also fixed a bug whereby setting constants on the FE didn't work in manual fitting.


cliffckerr:
works on my system but not robyn's, so closing for now

PR #655, opened 2016-02-03
Actually fix circumcision

cliffckerr:
@robynstuart please review. i think this works but i haven't tested it very much since i don't know how to adjust the cost curves. also please check the logic of the calculations of effmtct, effost etc. since these were all totally changed.


optima/model.py
```diff
@@ -69,12 +69,18 @@ def model(simpars=None, settings=None, verbose=None, benchmark=False, die=True):
     # Defined for total (not by populations) and time dependent [npts]
     treatvs     = simpars['treatvs']     # viral suppression - ART initiators (P)
     if usecascade:
-        biofailure  = simpars['biofailure']  # biological treatment failure rate (P/T)
-        vlmonfr     = simpars['vlmonfr']     # Viral load monitoring frequency (N/T)
-        restarttreat = simpars['restarttreat']  # Time to restart ART (T)
+        biofailure    = simpars['biofailure']  # biological treatment failure rate (P/T)
+        vlmonfr       = simpars['vlmonfr']     # Viral load monitoring frequency (N/T)
```
robynstuart: not a big deal, but i find this variable name very obtuse. how about freqvltest ?

optima/parameters.py
```diff
@@ -409,19 +410,23 @@ def makepars(data, label=None, verbose=2):
 
     # Aging transitions - these are time-constant transition rates
     duration = [age[1]-age[0]+1 for age in data['pops']['age']]
-    normalised_agetransit = [[col/sum(row)*1/duration[rowno] if sum(row) else 0 for col in row] for rowno,row in enumerate(data['agetransit'])]
+    normalised_agetransit = [[col/sum(row)*1.0/duration[rowno] if sum(row) else 0 for col in row] for rowno,row in enumerate(data['agetransit'])]
```
robynstuart: good one

PR #657, opened 2016-02-03
Bug/806 options for displaying curves and add notes not functional on cost function page

nanuxbe:


client/source/js/modules/project-set/cost-coverage/cost/cost.html
```diff
@@ -14,8 +14,8 @@
   <div ng-if="selectedProgram">
     <div style="display: flex; margin-top: 2%">
       <div class="chart-container __medium">
-        <div mpld3-chart="state.chartData" id="cost_cov_chart"></div>
-        <chart-toolbar chart-type="mpld3" mpld3-chart-data="state.chartData" chart-id="cost_cov_chart"></chart-toolbar>
+        <div mpld3-chart="state.chartData" id="cost_cov_chart" ng-if="state.chartData"></div>
```
jpuri: @nanuxbe: this chart tool-bar was a requirement in charts we should rather remove the other toolbar which appears when you hover. But I think that will be bigger change. This one should be good for now.

jpuri:
Thanks  a lot for the nice FE changes @nanuxbe :smile:

PR #658, opened 2016-02-03
API portions of Feature/731 1 5 3 0 5 add scenarios

hawkowl:


run.py
```diff
@@ -0,0 +1,42 @@
...
+base_resource = File('client/source/')
+base_resource.putChild('build', File('client/source/'))
+base_resource.putChild('api', OptimaResource(wsgi_app))
+
+site = Site(base_resource)
+
+endpoint = serverFromString(reactor, "tcp:port=8080")
+endpoint.listen(site)
+
+reactor.run()
```
nanuxbe: is this file part of the PR because it is supposed to or has it been added by mistake?

server/webapp/dbmodels.py
```diff
@@ -674,3 +682,65 @@ def recursive_delete(self, synchronize_session=False):
...
+            return op.Progscen(name=self.name,
+                               parsetname=parset.name,
+                               progsetname=progset.name,
+                               **self.blob)
+
+        elif self.scenario_type == "Parameter":
+
+            return op.Parscen(name=self.name,
+                              parsetname=parset.name,
+                              **self.blob)
```
nanuxbe: Hydrate ok, but I don't see any "restore" code. (it can be added as a separate card)

server/webapp/resources/project_scenarios.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+    method_decorators = [report_exception, login_required]
+
+    @swagger.operation()
+    @marshal_with(ScenariosDb.resource_fields, envelope='scenarios')
+    def get(self, project_id):
+        """
+        Get the scenarios for the given project
+        """
+        project_entry = load_project(project_id)
```
nanuxbe: `load_project(project_id, raise_exception=True)` makes the check for you and you can drop the next 2 lines

server/webapp/resources/project_scenarios.py
```diff
@@ -0,0 +1,196 @@
...
+    @swagger.operation(
+        parameters=scenario_parser.swagger_parameters(),
+        notes="""
+            Create a new scenario.
+
+            If it is a Parameter scenario, the request body should be a JSON
+            dict with the 'pars' key.
+        """
+    )
+    @marshal_with(ScenariosDb.resource_fields)
```
nanuxbe: if you use `responseClass=ScenarioDb.__name__` in `swagger.operation` you don't need to use `marshal_with (because there you are using the `resource_fields` on that class)

server/webapp/resources/project_scenarios.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        Create a new scenario for the given project.
+        """
+        args = scenario_parser.parse_args()
+
+        if args.get('scenario_type') not in ["Parameter", "Program"]:
+            raise ValueError("Type needs to be 'Parameter' or 'Program'.")
+
+        if args['scenario_type'] == "Parameter":
+            blob = {'pars': check_pars(json.loads(request.data))}
```
nanuxbe: why do you need to re-parse the request? Would something break if you were using `blob = {'pars': args}` ?

server/webapp/utils.py
```diff
@@ -204,6 +204,27 @@ def load_parset(project_id, parset_id, raise_exception=True):
     return _load_project_child(project_id, parset_id, ParsetsDb, ParsetDoesNotExist, raise_exception)
 
 
+def load_parset(project_id, parset_id, raise_exception=True):
```
nanuxbe: if you merge develop, you can refactor all those load methods and use `_load_project_child`https://github.com/optimamodel/Optima/blob/develop/server/webapp/utils.py#L170

nanuxbe:
All my comments are non-blocking and tests pass, so I'm merging this

PR #668, opened 2016-02-04
fix for saturation # display glitch in cost.html

boscoh:


robynstuart:
great, thanks @boscoh :+1:

PR #669, opened 2016-02-04
Constraints

cliffckerr:
implements constraints on optimizations and lots of other fixes


optima/optimization.py
```diff
@@ -58,25 +61,36 @@ def getresults(self):
...
     
-    Version: 2016jan26
+    Version: 2016feb03
     """
 
     printv('Defining default objectives...', 3, verbose=verbose)
+    
+    if type(progset)==Programset:
+        defaultbudget = sum(progset.getdefaultbudget()[:])
+    if type(project)==Programset: # Not actually a project, but proceed anyway
```
robynstuart: this is pretty confusing...

cliffckerr: agree, but just making it flexible, so you can do either `defaultobjectives(project)` or `defaultobjectives(progset)` since either will suffice

optima/optimization.py
```diff
@@ -58,25 +61,36 @@ def getresults(self):
...
+        defaultbudget = sum(project.getdefaultbudget()[:])
+    elif project is not None:
+        if progset is None: progset = 0
```
robynstuart: i think it's good to print a warning whenever this is done, but not urgent for now

optima/optimization.py
```diff
@@ -58,25 +61,36 @@ def getresults(self):
...
+        defaultbudget = sum(project.progsets[progset].getdefaultbudget()[:])
+    else:
+        defaultbudget = 1e6 # If can't find programs
```
robynstuart: as above

optima/optimization.py
```diff
@@ -96,18 +110,54 @@ def defaultobjectives(which='outcome', verbose=2):
     return objectives
 
 
+def defaultconstraints(project=None, progset=None, which='outcome', verbose=2):
+    """
+    Define default objectives for the optimization. Some objectives are shared
+    between outcome and money minimizations, while others are different. However,
+    outcome minimization is performed as part of money minimization, so it's useful
+    to keep all the keys for both. Still, ugly.
+
```
robynstuart: needs updating (again, not urgent, i am just nitpicking)

optima/optimization.py
cliffckerr: agree, verbose=2 or 3? i.e. show by default or only with extra output? probably 2?

optima/optimization.py
```diff
@@ -96,18 +110,54 @@ def defaultobjectives(which='outcome', verbose=2):
...
+    """
+    Define default objectives for the optimization. Some objectives are shared
+    between outcome and money minimizations, while others are different. However,
+    outcome minimization is performed as part of money minimization, so it's useful
+    to keep all the keys for both. Still, ugly.
+    
+    Version: 2016feb03
+    """
+
+    printv('Defining default objectives...', 3, verbose=verbose)
```
robynstuart: as above

optima/optimization.py
```diff
@@ -96,18 +110,54 @@ def defaultobjectives(which='outcome', verbose=2):
...
+    
+    if type(progset)==Programset: pass
+    elif type(project)==Programset: progset = project
+    elif project is not None:
+        if progset is None: progset = 0
```
robynstuart: warning here too please

optima/optimization.py
```diff
@@ -96,18 +110,54 @@ def defaultobjectives(which='outcome', verbose=2):
...
+        raise OptimaException('To define constraints, you must supply a program set as an input')
+
+    constraints = odict() # Dictionary of all constraints
+    constraints['name'] = odict() # Full name
+    constraints['min'] = odict() # Minimum budgets
+    constraints['max'] = odict() # Maximum budgets
+    for prog in progset.programs.values():
+        constraints['name'][prog.short] = prog.name
+        if prog.optimizable():
+            constraints['min'][prog.short] = 0.0
```
robynstuart: i wonder whether this should eventually be a property of programs? like how limits are a property of parameters. although, maybe that wouldn't work generally since programs are not fixed like parameters...

optima/optimization.py
```diff
@@ -126,16 +176,69 @@ def outcomecalc(budgetvec=None, project=None, parset=None, progset=None, objecti
...
         results.budgetvec = budgetvec # WARNING, not sure this should be here
         results.budgetyears = [objectives['start']] # WARNING, this is ugly, should be made less kludgy
-        results.budget = vec2budget(progset, budgetvec) # Convert to budget
+        results.budget = vec2budget(progset, budgetvec, optiminds) # Convert to budget
         return results
     else: 
         return outcome
 
 
+def constrainbudget(origbudget, total=None, limits=None, eps=1e-3):
```
robynstuart: should eps be from settings?

cliffckerr: it's a different eps since it's about a tolerance for a fit rather than to prevent divide-by-zero, plus settings isn't available here so i thought it might introduce an annoying and not really necessary dependency

cliffckerr: but good idea, i'll rename it to tolerance

optima/optimization.py
```diff
@@ -166,16 +269,39 @@ def minoutcomes(project=None, optim=None, inds=0, maxiters=1000, maxtime=None, v
...
+    
+    # Error checking
     if isnan(budgetvec).any():
-        budgetlessprograms = array(progset.programs.keys())[isnan(budgetvec)].tolist()
-        output = 'WARNING!!!!!!!!!! Not all programs have a budget associated with them.\n A uniform budget will be used instead.\n Programs with no budget are:\n'
-        output += '\n'.join(budgetlessprograms)
-        print(output)
-        budgetvec = zeros(nprogs)+totalbudget/nprogs
-    else: budgetvec *= totalbudget/sum(budgetvec) # Rescale
+        errormsg = 'Program "%s" does not have any budget' % progkeys[findinds(isnan(budgetvec))]
```
robynstuart: this would raise an error in cases where people had not specified a budget for a fixed-cost program, which may not be strictly speaking a problem for the subsequent steps.

optima/results.py
```diff
@@ -100,6 +101,46 @@ def __repr__(self):
         output += '============================================================\n'
         output += objrepr(self)
         return output
+        
+    
+    def __add__(self, R2):
+        ''' Define how to add two Resultsets '''
+        if type(R2)!=Resultset: raise OptimaException('Can only add two results sets')
```
robynstuart: i think you can actually add arbitrarily many resultsets, can't you? but you can only add a resultset to a resultset

cliffckerr: the __add__ method operates on two objects but can be done in succession, e.g. R1+R2+R3 = (R1+R2)+R3 etc.

robynstuart: i understand (that's how addition always works :) ) but i was suggesting that the exception should say something more like "Only resultsets can be added to resultsets."

cliffckerr: oh, got it :) will fix

robynstuart: ok. this is a nice method, anyway.

PR #670, opened 2016-02-04
feat(ProgramSet): make it non-mandatory to select population

jpuri:
@robynstuart, @guptaarun , @kitze , @nanuxbe plz review :)

Small change for card: https://trello.com/c/Sc2KtkKb/803-select-population-warming-message-is-shown-when-population-is-selected

### DIFF
--- a/client/source/js/modules/project-set/program-set/program-modal-ctrl.js
+++ b/client/source/js/modules/project-set/program-set/program-modal-ctrl.js
@@ -179,12 +179,6 @@ define(['./../module', 'angular', 'underscore'], function (module, angular, _) {
       });
     };
     $scope.formInvalid = false;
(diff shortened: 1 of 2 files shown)


robynstuart:
@jpuri i've looked at this, but the behaviour of the Add program still seems unpredictable. I don't know if we have time to fix all this now. For example, this workflow doesn't work as it should:
- Select "Edit" for a program
- Add a parameter
- Select a single population that applies to this parameter (not all populations).
- Save.
- Click "Edit" on the same program again. For the added parameter, "All populations" is selected in addition to the previously selected population

I can merge this branch but I'm concerned that the logic for editing programs seems to be a bit fragile. Can we address this after the workshop please?

PR #671, opened 2016-02-04
Bug/819 for autofit save doesn t work

nanuxbe:


jpuri:
FE changes look good @nanuxbe, just one console log is left :wink:

cliffckerr:
ready to merge

PR #672, opened 2016-02-04
Fix/use allpops instead of tot

jpuri:
Card: https://trello.com/c/oIY5vcQW/817-art-bug-in-cost-functions

@robynstuart , @guptaarun , @nanuxbe : plz review.


robynstuart:
@jpuri works great! thank you :+1:

PR #673, opened 2016-02-04
Bug/820 when multiple programs are targeting the same parameter only one program is displayed

nanuxbe:


robynstuart:
works great, thanks @nanuxbe !

PR #676, opened 2016-02-04
Temp benchmarking

cliffckerr:


cliffckerr:
just adding some extra benchmarks, since model got slow at several points

PR #680, opened 2016-02-04
Fix cascade 05

DrMud:
removing and shifting indicators from Cascade sheet


robynstuart:
thanks @DrMud all works fine now

PR #682, opened 2016-02-05
fixed parameter population selection bug

boscoh:


robynstuart:
works like a charm, thanks @boscoh !

PR #683, opened 2016-02-05
Fix/html error in effects page

robynstuart:
@boscoh please make sure i haven't done anything stupid here, and merge if not :)


boscoh:
Looks good to me.

PR #686, opened 2016-02-05
Feature/738 display results for scenarios

nanuxbe:


server/webapp/resources/project.py
```diff
@@ -951,3 +951,27 @@ def get(self, project_id):
...
+
+class Partnerships(Resource):
+    @swagger.operation(
+        summary="List partnerships for project"
+    )
+    @report_exception
+    @marshal_with(pship_fields, envelope="populations")
+    @login_required
+    def get(self, project_id):
+        project = load_project(project_id, raise_exception=True)
```
Ashalynd: Let's perhaps keep to the convention that xxx_entry is record and xxx_instance is hydrated thing?

nanuxbe: I've been using a lot of be_xxx we should clean that up after the workshop

PR #692, opened 2016-02-05
Fix cascade 06

DrMud:
Added propsupp back in but to Optional Indicators to use in 90-90-90 scenario. If available, it will override virallysupp which is based on viral monitoring and rebound.


optima/model.py
```diff
@@ -550,6 +552,12 @@ def model(simpars=None, settings=None, verbose=None, benchmark=False, die=False)
...
             ## Unsuppressed/Detectable Viral Load (having begun treatment)
+            currentusupp = people[usvl,:,t] # how many with suppressed viral load
+            currentsupp  = people[svl,:,t]
+            if propsupp[t]: # WARNING this will replace consequence of viral monitoring programs
+                currsupp  = currentsupp.sum(axis=0)
+                currusupp = currentusupp.sum(axis=0)
+                newsupptot = (propsupp[t]*currusupp - currsupp).sum()
             # 40% progress, 40% recover, 20% don't change cd4 count
             for cd4 in range(ncd4):
                 if cd4>0:
```
cliffckerr: on the next line, we should not have anything hard-coded -- these should be constants if appropriate

cliffckerr:
needs work before ready to merge

PR #694, opened 2016-02-07
Fix/coverage data optional

robynstuart:
Ensures that users can enter cost data without requiring coverage (necessary for fixed cost programs), and adds a method for getting default coverage, not currently used anywhere but probably useful

@cliffckerr please review.


optima/programs.py
```diff
@@ -193,15 +193,15 @@ def progs_by_targetpar(self, filter_partype=None):
...
+    def getdefaultbudget(self, t=None, verbose=2):
         ''' Extract the budget if cost data has been provided'''
         
         # Initialise outputs
         totalbudget, lastbudget, selectbudget = odict(), odict(), odict()
 
         # Validate inputs
-        if type(years) in [int, float]: years = [years]
-        if isinstance(years,ndarray): years = years.tolist()
+        if type(t) in [int, float]: t = [t]
```
cliffckerr: please use isnumber() from utils, because if e.g. `t` is type `numpy.float64` then this will fail

PR #698, opened 2016-02-07
Comparison table

cliffckerr:
adds method to programset to compare the currrent values from the parset to the current values from the budget. use doprint=True to print output, otherwise it returns it as a list of lists.

NB: we should think about whether or not it makes sense for a progset to have a 'parset' attribute. can you swap which parset you use with a given progset? if not, i think they should be linked.  then instead of

`comparison = P.progsets[0].compareoutcomes(parset=P.parsets[0], year=2016, doprint=True)`

the syntax would be

`comparison = P.progsets[0].compareoutcomes(year=2016, doprint=True)`

and likewise for getoutcomes etc.

@robynstuart please review.


robynstuart:
Closing as subsumed by PR699

PR #699, opened 2016-02-07
Autofit debugging

cliffckerr:
autofit actually works now :) added a doplot kwarg that will show a graph of what autofit is actually trying to minimize (nb, slows things down very substantially). also made results stored by name rather than by uid, although uid will work as well via getresults(). @robynstuart and/or @davidkedz please review. tests pass.

nb, this branch subsumes #698, so either review that one first, or review just this one and close/delete that one

### DIFF
--- a/optima/calibration.py
+++ b/optima/calibration.py
@@ -75,7 +75,7 @@ def sensitivity(project=None, orig=None, ncopies=5, what='force', span=0.5, ind=
 
 
 
-def autofit(project=None, name=None, what=None, maxtime=None, maxiters=1000, inds=0, verbose=2, doplot=False):
+def autofit(project=None, name=None, fitwhat=None, fitto=None, maxtime=None, maxiters=1000, inds=0, verbose=2, doplot=False):
     ''' 
     Function to automatically fit parameters.
     
@@ -96,7 +96,9 @@ def autofit(project=None, name=None, what=None, maxtime=None, maxiters=1000, ind
     parset.project = project # Try to link the parset back to the project -- WARNING, seems fragile
     origparlist = dcp(parset.pars)
     lenparlist = len(origparlist)
-    if what is None: what = ['force'] # By default, automatically fit force-of-infection only
+    if fitwhat is None: fitwhat = ['force'] # By default, automatically fit force-of-infection only
+    if type(fitwhat)==str: fitwhat = [fitwhat]
+    if type(fitto)==str: fitto = [fitto]
     if isnumber(inds): inds = [inds] # # Turn into a list if necessary
     if inds is None: inds = range(lenparlist)
     if max(inds)>lenparlist: raise OptimaException('Index %i exceeds length of parameter list (%i)' % (max(inds), lenparlist+1))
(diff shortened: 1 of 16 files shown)


cliffckerr:
nb, this branch subsumes https://github.com/optimamodel/Optima/pull/698, so either review that one first, or review just this one and close/delete that one

optima/calibration.py
```diff
@@ -181,7 +186,7 @@ def convert(pars, parlist, parvec=None):
     
 
 
-    def errorcalc(parvec=None, pars=None, parlist=None, project=None, bestindex=0):
+    def errorcalc(parvec=None, pars=None, parlist=None, project=None, fitto='prev', bestindex=0, doplot=False, verbose=2):
```
robynstuart: I think we need to rename this. There are many, many ways to quantify goodness of fit by some kind of "error calc". What you've got here looks like it's calculating the MAD/mean ratio or weighted absolute percent error. We could call this madmeanratiocalc or wapecalc, and then down the track we could consider adding other options (MAE, SMAPE, good old MSE).

cliffckerr: it sounds to me like then in future `errorcalc()` should take a `method` kwarg, to switch from MAD to MSE or something else -- how does that sound?

cliffckerr: i mean i could code up MSE right now, it'll take 2 min...

robynstuart: yes, kwarg would be great! and would you be against changing it to "gofcal" (for goodness of fit calc) or even objectivecalc? bc you're not actually calculating an "error" here.

cliffckerr: sure, will make those changes now

robynstuart: Eventually we could allow users could choose the GOF criteria, that would be amazing.

optima/calibration.py
```diff
@@ -224,14 +243,79 @@ def extractdata(xdata, ydata):
                         if nrows==1: modelrow = modelrows # WARNING, kludgy, should have the same shape!
                         else: modelrow = modelrows[row]
                         datax, datay = extractdata(results.datayears, datarow) # Pull out the not-NaN values
+                        if doplot: rowname = 'total' if nrows==1 else pars['popkeys'][row]
                         for i,year in enumerate(datax): # Loop over each data point available
+                            count += 1
                             modelx = findinds(results.tvec, year) # Find the index of the corresponding time point
                             modely = modelrow[modelx] # Finally, extract the model result!
-                            thismismatch = abs(modely - datay[i]) / mean(datay+eps)
+                            if method=='mad': thismismatch = abs(modely - datay[i]) / mean(datay+eps)
```
robynstuart: this isn't really the MAD, though - MAD would not normalise. how about weighted absolute percent error?

optima/calibration.py
```diff
@@ -224,14 +243,79 @@ def extractdata(xdata, ydata):
...
+            scatter(autofitresults['count'], autofitresults['mismatch'])
+            xlabel('Count')
+            ylabel('Mismatch')
+            ylim((0,ylim()[1]))
+            
+            subplot(rows,cols,2)
+            allmis = autofitresults['allmismatches'][-1] # Shorten
+            bar(arange(len(allmis)), allmis)
+            xlabel('Data point')
+            ylabel('Mismatch')
```
robynstuart: Can we please label it "Objective"?

optima/calibration.py
```diff
@@ -224,14 +243,79 @@ def extractdata(xdata, ydata):
...
+                    ylabel(key1+' - '+key2)
+                    ylim((0,ylim()[1]))
+            
+            show()
+            pause(0.001)
+            
+            
         
-        printv('Current mismatch: %s' % array(thismismatch).flatten(), 4, verbose=verbose)
+        printv('Current mismatch: %s' % array(mismatch), 4, verbose=verbose)
```
robynstuart: same here

optima/calibration.py
cliffckerr: i didn't really want to call it 'objective' because of possible confusion with optimization -- even just in terms of code maintainability i think it might end up causing problems to have two very different things share the same name. but i'm not wedded to 'mismatch' if there's another term that we could use.

optima/calibration.py
cliffckerr: ok, i'll rename and put actual mad in as an option too

robynstuart: :+1:

optima/calibration.py
robynstuart: you could label it with the method (mse/mad/etc)

cliffckerr: that's true, that would be a fair bit of extra work though, and on the FE users just see 'improvement', so would be inclined to leave it for now...

PR #700, opened 2016-02-07
Fix/scenario errors - Create program scenarios

guptaarun:
- Create program scenarios, expected BE GET and PUT to work with this payload https://jsonblob.com/56b5b639e4b01190df4d4de2
That is to avoid complexity in front-end code. 

@Ashalynd @robynstuart 

https://trello.com/c/HNM79o1e/733-2-3-1-create-program-scenarios


Ashalynd:
Please don't merge this, Jyoti is looking at it now. There was a misunderstanding regarding FE changes.

PR #706, opened 2016-02-08
feat(manage scenario): Validation for duplicate scenario name

guptaarun:
1) Prevent users creating two scenarios with the same name 
https://trello.com/c/FIyWYoQl/836-prevent-users-creating-two-scenarios-with-the-same-name

2) Hide 'run scenario' button if all scenarios are deleted from UI or no scenarios found then show only add buttons.


guptaarun:
@jpuri can you please review this PR.

jpuri:
Hey @robynstuart , can you plz look at functionality. I can check JS code.

nanuxbe:
Functionality works well. It might be useful to also check this on API level though

PR #722, opened 2016-02-09
Bug/846 year field in create program scenario pop up is to narrow

guptaarun:
- Create budget and coverage scenario year and program inputs field adjustments done
- Another issue fixed, when we try to remove year block earlier it just shifts the value to next block. Now it clears the input value based on removed year block index.


Ashalynd:
Looks good!

PR #727, opened 2016-02-10
Reconcile costfuncs with pars

cliffckerr:
this function works, and it's so good, omg. see testprograms.py and also testbest.py for example usage.


cliffckerr:
@robynstuart now ready to review and test, i think it's good to go

optima/programs.py
```diff
@@ -3,17 +3,18 @@
...
+from optima import OptimaException, printv, uuid, today, sigfig, getdate, dcp, smoothinterp, findinds, odict, Settings, runmodel, sanitize, objatt, objmeth, gridcolormap, isnumber, vec2obj, defaultrepr
+from numpy import ones, prod, array, shape, reshape, arange, zeros, exp, linspace, append, sort, transpose, nan, isnan, ndarray, concatenate as cat, maximum, minimum
 import abc
 
 # WARNING, this should not be hard-coded!!! Available from
 # [par.coverage for par in P.parsets[0].pars[0].values() if hasattr(par,'coverage')]
 # ...though would be nice to have an easier way!
-coveragepars=['numtx','numpmtct','numost','numcirc'] 
+coveragepars = ['numtx','numpmtct','numost','numcirc'] 
+programparsonly = ['numcirc'] # WARNING, maybe another way of handling this? Don't expect program and parameter values for numcirc to match
```
robynstuart: probably ok for now

optima/programs.py
```diff
@@ -375,7 +376,10 @@ def getoutcomes(self, coverage=None, t=None, parset=None, results=None, perturb=
 
         # Set up internal variables
         nyrs = len(t)
-        infbudget = odict((k,array([1e9]*len(coverage[k]))) if self.programs[k].optimizable() else (k,None) for k in coverage.keys())
+        try: infbudget = odict((k,array([1e9]*len(coverage[k]))) if self.programs[k].optimizable() else (k,None) for k in coverage.keys())
+        except:
+            errormsg = 'The following programs require coverage to be entered:\n%s' % '\n'.join(self.hasallcostcovpars(detail=True))
```
robynstuart: this isn't the right error message to display here. will amend.

optima/programs.py
```diff
@@ -556,79 +560,144 @@ def getpars(self, coverage, t=None, parset=None, results=None, ind=0, perturb=Fa
...
+        ## Store original values in case we need to go back to them
+        origvals = dcp(self.cco2odict(t=year))
+        pardict = dcp(origvals)
+        pararray = origvals[:] # Turn into array format
+        npars = shape(pararray)[0]
+        factors = ones((npars,1))
+        parlower = zeros(npars)
         
-        def vec2cco(): raise Exception('Not implemented')
+        ## Just do a simple random walk
```
robynstuart: not really sure how this is a simple random walk?

optima/programs.py
```diff
@@ -556,79 +560,144 @@ def getpars(self, coverage, t=None, parset=None, results=None, ind=0, perturb=Fa
...
+    ''' Calculate the mismatch between the budget-derived cost function parameter values and the model parameter values for a given year '''
+    factors = reshape(factors, (len(factors),1)) # Get it the right shape
+    pardict[:] = dcp(pararray * factors)
+    progset.odict2cco(dcp(pardict), t=year)
+    comparison = progset.compareoutcomes(parset=parset, year=year, ind=ind)
+    allmismatches = []
+    mismatch = 0
+    for budgetparpair in comparison:
+        parval = budgetparpair[2]
+        budgetval = budgetparpair[3]
```
robynstuart: ok for now but probably shouldn't be hard coded

cliffckerr: agreed

optima/programs.py
cliffckerr: hah, comment is outdated

Ashalynd:
@cliffckerr is this still relevant?

cliffckerr:
this branch? yes, this is the branch providing the BE for the reconcile() function :) sorry, we should've clarified that

robynstuart:
@cliffckerr
I ran reconcilepars in testprograms and saw that 
`before.covout['condcas'][('Clients', 'FSW')].ccopars = 
{'Condoms': [(0.35, 0.45)],
 'FSW programs': [(0.75, 0.85)],
 'intercept': [(0.2, 0.25)],
 't': [2016.0]}
`
while 
`after.covout['condcas'][('Clients', 'FSW')].ccopars = 
{'Condoms': [(0.24499999999999994, 0.315)],
 'FSW programs': [(0.78749999999999987, 0.89249999999999974)],
 'intercept': [(0.49499999999999988, 0.61874999999999991)],
 't': [2016]}
`

I thought that the idea was to choose parameter bounds within the ranges set by the user, but this method seems like it totally overwrites the values set by the user. I don't think this would be very nice behaviour - at the very least we should throw a warning if we do this, or better still we should constrain the algorithm to only search within the allowable ranges and return the best parameter values from there.

cliffckerr:
no, it's not even bounded by the absolute parameter limits, since it fits a multiplicative factor. definitely happy for it to lie within user-defined limits, but i do really think then that the user should be shown the (current conditions, current spend) point on some graph somewhere so they can see if the cost functions are consistent with parameter assumptions.

robynstuart:
and also, i ran it again and got 
`
after.covout['condcas'][('Clients', 'FSW')].ccopars
{'Condoms': [(1.4349999999999996, 1.845)],
 'FSW programs': [(0.74999999999999989, 0.84999999999999976)],
 'intercept': [(0.25999999999999995, 0.32499999999999996)],
 't': [2016]}
`
which is disconcertingly different from the previous time, also the values for Condoms shouldn't be greater than 1.

robynstuart:
oh sorry, i just saw your comment here. ok, so let's make it lie between user-defined limits and also show a graph/table.

PR #728, opened 2016-02-10
convert pop to tuple(str) as opposed to tuple(unicode)

boscoh:


optima/scenarios.py
```diff
@@ -141,6 +141,7 @@ def makescenarios(project=None, scenlist=None, verbose=2):
                     for pop in pops:
 
                         # Remove years after the last good year
+                        pop = tuple([str(p) for p in pop])
```
nanuxbe: Maybe this part of the code is not relevant to non-pship params. but shouldn't you check first if it's a single pop or pship?
this code will give things like `('M', 'S', 'M')` for single pops

cliffckerr: yes, i think you're right @nanuxbe

PR #731, opened 2016-02-10
fixed bug, ugg

cliffckerr:
fixes strange bug whereby diagnosed didn't show up in the cascade plot


cliffckerr:
would like review and would like tests to pass but also want to show the cascade :(

PR #734, opened 2016-02-10
(fix) revert tuple change in optima and so it in api

nanuxbe:


nanuxbe:
Cliff Kerr [4:52 PM] 
ok great, thanks @emma  , you can merge that to develop as far as i'm concerned, will be ready to deploy in ~10 min, will let you know

PR #738, opened 2016-02-10
Fix/add resultsref to parset

nanuxbe:


robynstuart:
thanks @nanuxbe looks good

PR #739, opened 2016-02-10
interpret "year" in program data as integer

Ashalynd:
@cliffckerr or @robynstuart feel free to review & merge

https://trello.com/c/64OKcsOK/865-year-returned-as-a-string-when-adding-data-to-programs


robynstuart:
@Ashalynd this looks as though it should work, but I've tried a couple of times to create a project and add a program with some cost data, but when i download the project it seems the year is still stored as a string. I'm not sure whether this is just something I'm doing wrong, so I'm going to merge the branch and see if it works for others, but I won't delete the branch in case more is required to get this fixed.

PR #742, opened 2016-02-11
Test optimization

cliffckerr:
makes optimization more robust


cliffckerr:
nvm, includes reconcile-prog-pars stuff which shouldn't be merged just yet

PR #745, opened 2016-02-13
updated README, and improved configuration

boscoh:
- added notes in readme on installing postgres properly
- added notes on configuring nginx
- fixed broken nginx.example.conf
- added notes on installing redis
- added a check in celery.sh for optima links in virtualenv
- fixed requirements check in run.sh and celery.sh


cliffckerr:
looks good, thanks for the additions!

PR #746, opened 2016-02-26
Treatment distribution

robynstuart:
Change the way that people move from diagnosed -> treatment so that those with lowest CD4 counts are taken first.

@cliffckerr or @davidkedz please review.


cliffckerr:
looks great, thanks!

PR #747, opened 2016-02-27
fixed anal transmissibility

cliffckerr:
fixed bug with insertive/receptive anal transmissibility constants in reverse order. @robynstuart please review.


robynstuart:
looks good, thanks!

PR #748, opened 2016-03-03
Cleaned up imports

cliffckerr:
This should be the "correct" way of doing imports so it would be really nice to implement this... @robynstuart let's figure out why it didn't work before on your computer and maybe @boscoh you could check to make sure it works on yours


boscoh:
This didn't work for me either.

I noticed that the modules within optima import from a global optima object. This only works if you've somehow bootstrapped optima into the global python site-packages on your local machine. The correct way is not to use optima but to import from the exact sibling.

cliffckerr:
and now even i can't figure out how i got this to work before :/

cliffckerr:
wait, never mind, if you clear out the python cache -- e.g. remove `optima.pyc` -- it should work

PR #750, opened 2016-03-04
Convert 1.0 projects

cliffckerr:
@robynstuart do you want to merge so people can start playing around with this? it may not be 'done' done for a long time but is already usable


cliffckerr:
i removed all the legacy code in this branch too -- if you want could make that a separate PR

PR #751, opened 2016-03-04
Dont reconcile pop sizes

cliffckerr:
@robynstuart please review. i think this is the better way of handling it. it checks to see if there are any in flows (from another age group or births) and if not, then forces the pop size to match that specified. otherwise, it calculates it. seems legit. btw, i apologize for the title of this PR. apostrophes are hard to do from the command line (https://xkcd.com/1638/).


robynstuart:
seems to work, thanks!

PR #752, opened 2016-03-10
Expanded Text entry in Scenario Parameter Mode

boscoh:


cliffckerr:
looks good thanks @boscoh !

PR #753, opened 2016-03-11
Fix prop params

cliffckerr:
@robynstuart it works; it's not perfect. it introduces some warnings in parameters.py since now using nan in a comparison, but these are harmless.


robynstuart:
:+1:

PR #756, opened 2016-03-15
FE fixes

boscoh:
- widens the underlying matplotlib graph to include plenty of room for legends
- in optima/plotting, tweaked the legends and added humanized y-axis (however doesn't propagate to web)
- added a check for the optima.pyc in run.sh
- improved osx_boot_env.sh
- fixed the wrongful conversion to ints for unit-cost in cost-coverage curve
- fixed an error with multiple years in cost-coverage curve
- improved tables in cost-coverage graph page

### DIFF
--- a/optima/plotting.py
+++ b/optima/plotting.py
@@ -25,30 +25,30 @@ defaultplots = ['improvement', 'budget'] + defaultepiplots # Define the default 
 
 
 
-def humanise_y_ticks(ax):
+def humanizeyticks(ax):
     vals = list(ax.get_yticks())
-    max_val = max([abs(v) for v in vals])
-    if max_val < 1e3:
+    maxval = max([abs(v) for v in vals])
+    if maxval < 1e3:
         return map(str, vals)
-    if max_val >= 1e3 and max_val < 1e6:
+    if maxval >= 1e3 and maxval < 1e6:
         labels = ["%.1fK" % (v/1e3) for v in vals]
-    elif max_val >= 1e6 and max_val < 1e9:
+    elif maxval >= 1e6 and maxval < 1e9:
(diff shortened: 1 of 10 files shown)


server/run.sh
```diff
@@ -12,15 +12,19 @@ if [ ! -f "config.py" ]; then
   cp config.example.py config.py
 fi
 
+# delete to avoid some module import problems
+if [ -f "../optima/optima.pyc" ]; then
+  rm ../optima/optima.pyc
```
cliffckerr: this was a problem due to a one-off change, so i'm not sure we need to enshrine it in code...although if we do, shouldn't we just remove all .pyc files?

optima/plotting.py
```diff
@@ -25,6 +25,40 @@
 
 
 
+def humanise_y_ticks(ax):
```
cliffckerr: i know we all have different conventions, but i would prefer to stick to a single convention within a project -- in optima so far we've been using US spellings and no underscores -- obviously for autumn and nutrition happy to comply with the existing convention for those

server/run.sh
boscoh: I put this in line because in one of the recent pulls, I got the problem again. It's a real head-scratcher because the error gives no indication of the problem. By having it in the script, I hope it will eliminate the problem for other people as well. My suggestion is to leave it in for a while, and then take it out when most people's systems are clear. 

Also, the idea of removing all .pyc is an excellent one because I remember many bugs in the past related to invalidated caches.

optima/plotting.py
boscoh: good point. will fix. 

i have to pay extra attention as alas, there are different coding conventions within the javascript modules, in the server code, the optima code, and mine. :(

PR #763, opened 2016-03-24
added functions for cleaning results out of a project

cliffckerr:
i was more than a little traumatized to see some single project files pushing 300 mb :) this PR introduces a new save() method for projects that by default doesn't save results; it also introduces a cleanresults() utility that acts on a single file or on all files in the current directory; and lastly it makes it so the default projects can actually be loaded from any directory rather than just the optima or tests directories.

### DIFF
--- a/optima/__init__.py
+++ b/optima/__init__.py
@@ -68,7 +68,7 @@ from .colortools import alpinecolormap, bicolormap, gridcolormap, vectocolor
 
 ## Utilities
 from . import utils # Load high-level module as well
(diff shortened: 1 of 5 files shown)


cliffckerr:
btw can be tested with e.g.
```
import optima as op
P = op.defaults.defaultproject('best')
P.runsim()
P.save(saveresults=True)
op.cleanresults()
```

davidkedz:
Project files finally have a save method? Aww sweet! Merged.

PR #767, opened 2016-04-06
Colombia fixes

cliffckerr:
Added `usemeta` kwarg so `interp()` can be used to get new `y` parameter values


robynstuart:
bueno mate

PR #768, opened 2016-04-07
Massive cleanup of FE graphing code

boscoh:
- deleted all unused graphing code ~ 30 files
- moved all graphing files into `mpl3d-charts'
- removed all cut-and-paste javascript code for graph reformatting from all HTML templates 
- do graph reformatting at the single point of graph injection (doesn't require polling)
- added 10k, 10m, 10b to tick labels
- tweaked better formatting on graph pages -> graphs should look reasonably grid-like now


cliffckerr:
this looks amazing!! i always wondered why things needed to be so complicated -- and you showed that they didn't :) will test soon.

cliffckerr:
all works, looks great!!

PR #769, opened 2016-04-07
Doc fixes and Costcov doc and small Costcov fix

boscoh:
- improved readme especially regards to the database set up
- added a fix to get FE unit-testing working
- started making the FE a relative URL-based system with the intent of removing NGINX dependency for local web-version
- documented Costcov objects
- added a fix for certain array errors when using Costcov
- improved the scripts for starting server


server/sandbox.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
cliffckerr: why is this called `sandbox.sh`? looks like it should be called `env.sh` or something...to me sandbox implies testing code, not production code

optima/programs.py
```diff
@@ -16,6 +16,18 @@
...
 
+
+def promotetoarray(l):
+    if isnumber(l):
+        return array([l])
+    elif isinstance(l, list) or isinstance(l, tuple):
+        return array(l)
+    elif isinstance(l, ndarray):
+        return l
+    raise OptimaException("Expecting a number/list/tuple/ndarray sequence: %s" % repr(l))
```
cliffckerr: for clarity might be preferable to have an explicit `else` here

server/sandbox.sh
boscoh: I'm happy to change it. For me tho, sandboxing is a general concept - for instance, one reason chrome uses so much memory is because it sandboxes every page from each other.

optima/programs.py
boscoh: no problems with that. in future, I'd probably centralise error-checking like this throughout the codebase.

server/env.sh
```diff
@@ -0,0 +1,40 @@
...
+sort ./requirements.txt > ${TMP_DEPS}.requirements.txt
+if ! cmp ${TMP_DEPS}.requirements.txt ${TMP_DEPS} > /dev/null 2>&1
+then
+  echo "Installing Python dependencies ..."
+  cat ${TMP_DEPS}
+  pip install -r ./requirements.txt
+fi
+
+# Ensure that the optima modelling module is available from the sandbox
+if [ ! -f "./p-env/lib/python2.7/site-packages/optima.egg-link" ]; then
```
cliffckerr: i'm not sure this is right -- wouldn't you need `pip install -e PATH` to install it here? so won't this block always run?

server/env.sh
```diff
@@ -0,0 +1,40 @@
...
+    echo "Installing optima in virtualenv for the celery webapps..."
+    cd ..
+    python setup.py develop
+    cd server
+fi
+
+# delete to avoid some module import problems
+if [ -f "../optima/optima.pyc" ]; then
```
cliffckerr: this should be all .pyc files or none, i think...in fact let's just kill all of them

server/run.sh
```diff
@@ -1,38 +1,12 @@
 #!/bin/bash
 
-if [ ! -d "./p-env/" ]; then
-  if [ "$1" == "--system" ]; then
-    virtualenv --system-site-packages p-env
-  else
-    virtualenv p-env
-  fi
-fi
+source ./env.sh
```
cliffckerr: @boscoh without `source` this doesn't work for me -- wrong paths i think

PR #775, opened 2016-04-11
fix-odict-repr

cliffckerr:
fixes repr problem for ipython for odicts, and generally makes reprs nicer (juicy information at bottom instead of top), and adds `disp()` method to odicts which prints with one entry per line (by default), etc.

@robynstuart @boscoh please review. *NB* you will need to delete your *.pyc files, or else all hell will break loose.


robynstuart:
works fine now, thanks

PR #778, opened 2016-04-15
FE Improvement to allow for Project Upload 

boscoh:
Project Upload page:
- will check for names that already exist
- automatically fills in name with filename + variant if the name is empty
- automatically open uploaded project

Programset page:
- Can now display the programs of an uploaded project properly
- big cleanup of program reading/writing code in the db to allow for this

New style entry-table:
- a new generic data-entry table that can be reused across pages
- entry inputs always available, old entries can be edited
- clean compact look
- allow selectors for things like years, and programs
- allow adaptable selectors such as estimated population in cost-coverage table

Cost coverage page:
- the first option is automatically selected for all drop-down menus
- ticks on graphs now shows 'k', 'm', 'b' properly
- cleaned up API, uses already uploaded data to construct tables, rather than more API calls
- all writes go to a single program write API call
- simplified the popsizes api call to get populations to estimate popsizes
- DB updates occurs on accepting changes with ticks; no need for separate update button clicks

Chart controls:
- PNG and SVG button, no popups
- in scenarios: all graphs in the backend have been merged into a single angular directive
- consolidated a lot of charting code into a single module

Calibration:
- massively refactored parset extraction into JSON objects - code is much cleaner
- all values are converted to float before being sent to the DB and to the simulation - no more lost significant figures
- auto-fit selects 10 sec by default
- polling output per second to provide feedback

Cost outcome page:
- designed a more useful entry-table -> still requires a bit of thinking and checking, but should be more useful, both to scan and edit.

Scenarios page:
- no unncessary popups
- defaults are selected
- a default scenario name is generated
- data is presented in a table format
- data entry is ready on intializations
- run scenarios generates graph using the default selections of graphs

However, for scenarios page:
- only parameter scenarios work
- there's an issue with target populations condcas etc that I have to understand
- budget and coverage values can be entered but the server code rejects - I have to understand how to work out when these values can be entered
- add a selection bar for graphs - basically figure out how selections work


robynstuart:
@boscoh this is great, thank you!

I can merge this now so that you've got a basis to work from for today. But there are a few things that will need to be fixed in a future PR. The point in bold is the highest priority:

Project upload
1. Projects should be auto-named without the .prj extension

Program set page
**1. Something is going wrong with programs not being saved to program sets.** I created a program set for a newly-created project, selected one program as active and saved it. But then when I navigated to the cost-cov page, there were no programs in the drop-down list. And when I navigated back to the program set page, the program was not longer marked as active for the program set.

Cost-coverage page
1. No longer displays "Options for displaying curves"
2. No longer displays "Add notes" section
3. No longer automatically scales page contents to fit screen
4. Ticks on y axis of graph shows 'k', 'm', 'b' properly, but not on x axis 

Scenarios page
1. Table for adding parameters to parameter scenarios does not auto-scale
2. I still find the display of results for the scenarios difficult to navigate & use, but it sure is nice to see graph selectors there :) Still, we should do something about the fact that 3/4 of the screen is taken up with the Optima header, the scenario control panel and the graph selector panel.

boscoh:
Thanks. Robin. I’ve typed up your comments as github issues: https://github.com/optimamodel/Optima/issues.

A quick note about the notes and options in the cost-coverage page - they weren’t actually being saved to the DB, so I switched them off, and will switch them back on when the DB is written.

Bosco
http://boscoh.com


On 12 May 2016 at 8:42:54 AM, robynstuart (notifications@github.com) wrote:

@boscoh this is great, thank you!

I can merge this now so that you've got a basis to work from for today. But there are a few things that will need to be fixed in a future PR. The point in bold is the highest priority:

Project upload
1. Projects should be auto-named without the .prj extension

Program set page
1. Something is going wrong with programs not being saved to program sets. I created a program set for a newly-created project, selected one program as active and saved it. But then when I navigated to the cost-cov page, there were no programs in the drop-down list. And when I navigated back to the program set page, the program was not longer marked as active for the program set.

Cost-coverage page
1. No longer displays "Options for displaying curves"
2. No longer displays "Add notes" section
3. No longer automatically scales page contents to fit screen
4. Ticks on y axis of graph shows 'k', 'm', 'b' properly, but not on x axis

Scenarios page
1. Table for adding parameters to parameter scenarios does not auto-scale
2. I still find the display of results for the scenarios difficult to navigate & use, but it sure is nice to see graph selectors there :) Still, we should do something about the fact that 3/4 of the screen is taken up with the Optima header, the scenario control panel and the graph selector panel.

PR #799, opened 2016-05-17
Restore manualfitlist

boscoh:


optima/parameters.py
```diff
@@ -923,7 +923,183 @@ def listattributes(self):
...
+            vtype = type(tmppars[key].y)
+            tmppars[key].y = vtype(value)
+            printv('%s.y = %s' % (key, value), 4, verbose)
+        else:
+            print('Parameter type "%s" not implemented!' % ptype)
+
+            # parset.interp() and calculate results are supposed to be called from the outside
+
+
+def manualfitlists(self, ind=0):
```
cliffckerr: um, @boscoh this is in here twice :)

cliffckerr:
still not ideal -- the 'reset' button is broken -- but everything else seems to work

PR #801, opened 2016-05-18
Improved handling of program sets & working outcome editing page

boscoh:
This is another large update:
- worked on DB loading/saving of new programs, and integrates with programs loaded from .prj
- completely redesigned the outcomes page

-


robynstuart:
@boscoh: wow, massive effort again, thanks!

I've reviewed this only from a functionality point of view, I'm not up to the task of checking all the code changes. Can you please (1) look at my functionality comments, and then (2) start working with @critcortex - I suggest you do this by showing her what you have done here so that she can then review the DB changes?

Functionality comments:
Ok, so I really don't think that the new table format on the outcome page is working... Can we please revert to the old one? This one has several issues, namely:
1. It doesn't display the correct information. To test it, I created a single program, which affects one parameter. But the table displayed asks me to enter data for 2 programs. This shouldn't be possible. In the old version, when you selected a parameter the table would automatically generate rows with entry fields for each program that targeted that parameter (1 program, 1 row; x programs, x rows, etc). 
2. The old table had useful text that was intended to help people fill out the table, which this new table is lacking. Users will not know how to fill in values for "Lo" and "Hi". 
3. The table doesn't fit nicely onto the page.
4. The fields don't validate the entries. It shouldn't be possible for users to enter negative numbers for condom use, for example. The old table had field validation, and furthermore, the permissible values for each field were displayed above the fields.

All in all, I much preferred the old version. I know that it'd be nice to have a single table that we could use on every page, but I don't think it's going to work since all the pages require different formats of information.

PR #1727, opened 2017-10-26
fix-disappearing-target-population

cliffckerr:
@robynstuart seems to fix the problem, although i did this by removing a `deepCopy()`, which i assume was there to prevent some _other_ bug, so i am a little concerned about what the consequences might be. haven't found any way to make this break, though.

also removed a pointless FE restriction that programs targeting a `tot` parameter had to target all populations (which isn't true for e.g. OST).

running on sandbox.

cc @gchadder3, maybe spend 10 min looking at `programs-modal.js` and make sure i didn't do anything obviously incompetent by removing the `deepCopy()`. subtly incompetent is fine though.

### DIFF
--- a/client/source/js/modules/programs/program-modal.js
+++ b/client/source/js/modules/programs/program-modal.js
@@ -26,7 +26,7 @@ define(['angular', 'underscore'], function (angular, _) {
       $scope.state = {
         selectAll: false,
         isNew: !program.name,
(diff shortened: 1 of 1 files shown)


cliffckerr:
still a bug:
* edit target populations
* click save
* go to another page
* they aren't saved

PR #1730, opened 2017-10-29
Time varying optimization

cliffckerr:
@robynstuart Here is time-varying optimization. Not implemented yet on the FE (although that should be pretty straightforward), and still need to iron out bug https://trello.com/c/lsBgRdSK/900-issue-1696-outcome-mismatch which is causing problems when you try to run it with `best` instead of `simple`, but you can start reviewing if you want.

Note that it can be run either with the total budget constrained in each year (`tvconstrain=True`), or keeping the total the same across all years but letting it change year by year (`tvconstrain=False`)

Example code:

```python
from optima import demo, defaulttvsettings, outcomecalc, pygui
optimize = 0 # Either run an optimization, or an outcome calculation
P = demo(0, which='simple')

if optimize: 
    tvsettings = defaulttvsettings(tvconstrain=False)
    P.optimize(timevarying=1, mc=0, maxiters=20, randseed=1, verbose=2, tvsettings=tvsettings)
    pygui(P)

else:
    args = {'project':       P, 
            'budgetvec':     P.progset().getdefaultbudget(),
            'tvsettings':    defaulttvsettings(tvconstrain=True),
            'outputresults': True}
    result = outcomecalc(tvcontrolvec=[-5,0], **args)
    pygui(result)
```


robynstuart:
@cliffckerr I got an exception when running TV optimization with the concentrated demo project on sandbox. 
```
 File "/home/optima/installations/sandbox/optima/optimization.py", line 283, in constrainbudget
    raise OptimaException(errormsg)
OptimaException: final budget amounts differ (56118584.070796 != 67800000.000000)
Optima debugging info:
   Version: 2.7
   Branch:  time-varying-optimization
   SHA:     2feda78
   Path:    /home/optima/installations/sandbox
```

PR #1731, opened 2017-11-01
Plot all cascade bars

robynstuart:
Plots a full cascade rather than a condensed version. Required for ongoing HIV applications.

@gchadder3 could you please review & merge (since CK on leave)?

Let me know if any questions.


gchadder3:
I ran `testbest.py` but am not sure whether the bars that are coming up for the "Care cascade" plot are what are desired.  Both of the sets of vertical colored bars (purple and green) are of equal height, and both are for the same year (2015).  I wonder if that was desired or whether the purple bars ought to be for the start year (2000).

PR #1746, opened 2018-04-02
Include saturation

cliffckerr:
@robynstuart i think there was a bug so that saturation wasn't being counted in program interactions -- i.e., if you set saturation to 10% and the budget was enough to cover 10% of the population, then for the sake of the covout function, coverage was being counted as 100% (instead of 10%). i think this fixes it. pls test and review. related to card https://trello.com/b/CSqYTPJf/optima-hiv-development

### DIFF
--- a/optima/programs.py
+++ b/optima/programs.py
@@ -385,10 +385,10 @@ class Programset(object):
         return thisbudget
 
 
-    def getdefaultcoverage(self, t=None, parset=None, results=None, verbose=2, sample='best'):
+    def getdefaultcoverage(self, t=None, parset=None, results=None, verbose=2, sample='best', proportion=False):
         ''' Extract the coverage levels corresponding to the default budget'''
         defaultbudget = self.getdefaultbudget()
-        defaultcoverage = self.getprogcoverage(budget=defaultbudget, t=t, parset=parset, results=results, sample=sample)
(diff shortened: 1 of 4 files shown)


optima/programs.py
```diff
@@ -548,7 +548,8 @@ def getoutcomes(self, coverage=None, t=None, parset=None, results=None, sample='
                             if not self.covout[thispartype][thispop].ccopars[thisprog.short]:
                                 print('WARNING: no coverage-outcome parameters defined for program  "%s", population "%s" and parameter "%s". Skipping over... ' % (thisprog.short, thispop, thispartype))
                                 outcomes[thispartype][thispop] = None
-                            else: outcomes[thispartype][thispop] += thiscov[thisprog.short]*delta[thisprog.short]
+                            else: 
+                                outcomes[thispartype][thispop] += thiscov[thisprog.short]*delta[thisprog.short] * self.programs[thisprog.short].costcovfn.getccopar(t=t, sample=sample)['saturation']
```
robynstuart: @cliffckerr I don't think this is right. Saturation is supposed to enter in via the cost-coverage function, so it should already have been taken into account and shouldn't need to come in again via the coverage-outcome function. I ran `testmodalities` and the test there doesn't pass with this line here (it does pass if I revert to the previous line).

cliffckerr: @robynstuart perhaps the fix isn't correct, but the current version is _definitely_ wrong: see `analyses/debug/modalities.py`, which demonstrates that if you reduce the saturation, you _increase_ the impact of the program.

PR #1747, opened 2018-04-12
Diagnosed births

robynstuart:
Previously, all HIV+ births were being assigned to the undiagnosed compartment, even if the mother was diagnosed. This fixes that.

@cliffckerr if you could review as soon as possible, please, as this is needed for Swaziland (cc @sherriekelly)


cliffckerr:
@robynstuart please increment the version number if you make a change to how the model runs, so running the same model number will always produce exactly the same result

PR #1748, opened 2018-04-25
Ban f circ

robynstuart:
Addressing the issue noted by Zara, namely that it's possible to set >0 values for the proportion of people circumcised among female populations.

@cliffckerr please review. If it's not complicated, we could additionally prevent the FE from displaying female populations associated with the `proprcirc` parameter. However, I think it's probably sufficient that the model just fails.


cliffckerr:
superseded by `fix-circumcision-bug`

PR #1754, opened 2018-04-30
Age varying mortality

robynstuart:
@cliffckerr please review

Adds a parameter to control HIV death rates by population, as requested by WB. (see https://trello.com/c/KEM2koVP/998-age-specific-mortality-rates).


cliffckerr:
@robynstuart another way to handle this would be to include it as a `Timepar`, which would allow variation over both age and time (it could be entered in the databook just under the background mortality rate table). it's a bit messy either way, because people aren't used to entering metaparameters in the databook...

another option would be to introduce additional constants for defined age bands. then it should be fairly straightforward to decide which mortality rate to use for each population the user has entered.

i don't have a strong preference, and probably have a slight preference for the way you've implemented it in this PR, but it might be worth considering these options...

cliffckerr:
@robynstuart ok, sounds good, will review & merge as-is

PR #1757, opened 2018-05-02
Demo fails gracefully

cliffckerr:
@robynstuart simple change so that for a new user, if they haven't set up the `_OptimaDemo` account, it shows a blank drop-down rather than crashing. not easy to demo on sandbox so haven't set it up there, but i think you can just merge...


robynstuart:
ok

PR #1758, opened 2018-05-03
Remove 'estimated population' column from cost functions page

mariadelmarq:
Removed the 'estimated population' column from cost functions page and updated the package.json file so that 'npm install' works on Windows.


cliffckerr:
looks good thanks @mariadelmarq !

PR #1763, opened 2018-05-21
Fix/mpops for progs

robynstuart:
@cliffckerr this could be improved, but it's an improvement on the current state of things. See report here: https://trello.com/c/QmZyp8VJ/1023-issue-with-creating-vmmc-program. This PR is aiming to address this and it addresses issues #1 and #3 on the card, but not the issue whereby it's still possible to select female pops here.

Running on sandbox, test by loading the spreadsheet on the trello card.


cliffckerr:
@robynstuart i can see how to fix this properly so only male populations appear, but it's quite a bit of work and would take a lot of testing to make sure it didn't break anything else, so i would like to leave it as a known issue for now.

PR #1764, opened 2018-05-21
Tv treatfail

robynstuart:
Introduces new logic for handling the relationship between treatment failure, viral suppression, and VL monitoring, as per https://trello.com/c/HnrKvBxp/1009-consider-changes-to-cascade-logic

Main changes are:
1. Adds a time-varying treatment failure rate as per https://trello.com/c/WApncXdQ/1008-time-varying-treatment-failure-rate
2. Adds a new parameter for the proportion of people who are switched to a new regimen after having been discovered to be failing (DW request)
3. Makes it possible for people to switch to using the proportions entered in the optional indicators sheet, i.e., if they want to treat them as actual parameters instead of things to calibrate to.

Note, I added a scenario to investigate the impact of the first two parameters, see `testscenarios.py`

@cliffckerr please review and share your thoughts. I consider this WIP but ready for input.


optima/parameters.py
```diff
@@ -126,14 +126,43 @@ def fixprops(self, fix=None, which=None, startyear=None):
...
+        if startyear is None:
+            if fix:  startyear = self.pars['numtx'].t['tot'][-1]
+            else:    startyear = 2100
         if 'dx'   in which: self.pars['fixpropdx'].t   = startyear
         if 'tx'   in which: self.pars['fixproptx'].t   = startyear
         if 'supp' in which: self.pars['fixpropsupp'].t = startyear # Doesn't make sense to assume proportion on treatment without assuming proportion suppressed....also, crashes otherwise :)
         return None
         
     
+    def usedataprops(self, use=None, which=None):
```
cliffckerr: @robynstuart i don't know about this...i think we should instead have this in the `Testing & treatment` sheet maybe

optima/parameters.py
```diff
@@ -126,14 +126,43 @@ def fixprops(self, fix=None, which=None, startyear=None):
         if   which is None:  which = ['tx','supp']
         elif which is 'all': which = ['dx','tx','supp']
         else:                which = promotetolist(which)
-        if fix:  startyear = self.pars['numtx'].t['tot'][-1]
-        else:    startyear = 2100
+        if startyear is None:
```
cliffckerr: wait, isn't this a different `startyear`? i think this is a bug and should be a different variable name entirely

optima/parameters.py
robynstuart: Let's follow up over email

robynstuart: @cliffckerr I believe we resolved this and are going to keep this - ok?

optima/parameters.py
robynstuart: I don't think it's a different `startyear`? It looks like I should be able to call thie method with a startyear, and in the previous implementation, it wouldn't be used...

robynstuart:
@cliffckerr thanks for waiting as I made final changes after following up with Clemens and others. I think this is ready now, can you pls re-review and merge if satisfied?

cliffckerr:
@robynstuart looks good, no changes!

PR #1768, opened 2018-05-29
Refresh matrices

robynstuart:
To fix https://trello.com/c/9FVJNSmZ/1026-refreshparset-does-not-update-risktransit


robynstuart:
Merging due to urgency

PR #1769, opened 2018-05-31
Better spreadsheet errors

cliffckerr:
@robynstuart changes:

* checks the version of the spreadsheet to match optima; prints a BE warning if it doesn't match, or, if an exception is raised, prints the version mismatch along with the exception
* actually checks to make sure all parameters have been populated, and prints out if any haven't
* removes "pops" and "meta" from `model-inputs.xlsx` and instead just loads these explicitly (so now everything in there is actually a data input)
* gives a more helpful warning if too many rows, rather than just "pop from empty list"

not tested super extensively, but tests pass, and seems to give at least _better_ errors with common spreadsheet errors (e.g. new or removed parameters).


robynstuart:
looks good, thanks @cliffckerr

PR #1773, opened 2018-06-07
Grab bag o fixes

cliffckerr:
Fixes a bunch of random long-standing bugs, including
* https://trello.com/c/nSYcfj2F/1029-calibration-refresh-text-edits -- simple text changes as suggested by Sherrie
* https://trello.com/c/VKAIASiM/1019-issue-with-graph-dl-under-advanced-options -- Bosco was passing "advanced" as a plot selection and then removing it in some places but not others
* https://trello.com/c/ohm2piLZ/931-copy-projects-can-break -- this is the main change, it is much more accepting of broken links to parsets. It won't show results from a since-deleted parset, but it won't give a stack trace either (it will give a stack trace if you try to run an optimization with a deleted parset though).

@robynstuart pls review, now running on sandbox.

### DIFF
--- a/client/source/js/modules/calibration/calibration.js
+++ b/client/source/js/modules/calibration/calibration.js
@@ -414,10 +414,10 @@ define(['angular', 'underscore'], function (angular, _) {
       modalService.choice(
         refreshparsetandprev,
         refreshparsetwithoutprev,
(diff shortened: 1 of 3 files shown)


robynstuart:
look good, thanks

PR #1774, opened 2018-06-08
Count time up from zero

mariadelmarq:
I implemented the timer in an alternative way, to achieve a solution for this Trello card: https://trello.com/c/DiFREYIk/1016-fe-optimizations-time-running-counts-up-from-negative-numbers. This timer is used for calibrations and optimizations.


cliffckerr:
looks beaut, thanks @mariadelmarq !

PR #1775, opened 2018-06-20
Remove years from plotcascade function call

mariadelmarq:
The cascade bars should not be plotted between the simulation years, the function already handles the default values for years depending on the data and the simulation (lines 894 to 903).

Solves Trello card: https://trello.com/c/X9mdsMTE/966-cascade-plots-are-broken-on-fe.


mariadelmarq:
Can someone please review? Thanks!

PR #1776, opened 2018-07-12
Change ART projection assumption from buttons to radio selectors

mariadelmarq:
Reading in what the value is set to for each parameter set to determine initial state.

Solves Trello card: https://trello.com/c/mU3OSdAu/1006-change-proportion-art-button-to-radio-selector

@robynstuart and/or @cliffckerr : please triple check my Python code commits! :)


cliffckerr:
@mariadelmarq works like a charm!

PR #1777, opened 2018-07-12
Enable multi-select for upload projects and spreadsheets

mariadelmarq:
Solves Trello card: https://trello.com/c/e3uvTlOp/932-enable-multi-select-for-upload-projects.


cliffckerr:
@mariadelmarq it works, the only issue is that if you upload two projects with the same name at the same time, they'll still have the same name...but that's such a corner case that i don't think it's worth worrying about

PR #1778, opened 2018-07-12
Correct optimization name

mariadelmarq:
Use the optimization name that corresponds to the delete and copy buttons, rather than the one that is currently selected.

Solves Trello card: https://trello.com/c/nwL2fCh6/958-fe-sandbox-when-clicking-on-the-delete-button-for-optimizations-it-always-asks-are-you-sure-you-want-to-permanently-remove-optim.

@cliffckerr and/or @robynstuart : I checked the edit and download buttons, and they're right. I wasn't able to confirm the upload button, though...


cliffckerr:
@mariadelmarq looks good, thx!

PR #1779, opened 2018-07-13
Respect selected plots

mariadelmarq:
Change default plots for "Advanced options": remove "xxx - total" and "New HIV infections caused - xxx" and respect selection of graphs and zoom when toggling between "Normal options" and "Advanced options", as well as when switching tabs ('Calibration', 'Projects', 'Programs', etc). Previous behaviour was to reload default graphs for each each time the options or tabs were toggled.

Solves Trello card: https://trello.com/c/ISDj8EBn/1033-calibration-advanced-display-options-update-default-maintain.

### DIFF
--- a/client/source/js/modules/charts/charts.js
+++ b/client/source/js/modules/charts/charts.js
@@ -402,20 +402,21 @@ define(
             });
         };
 
-        scope.switchGraphs = function() {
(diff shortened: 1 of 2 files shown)


cliffckerr:
@mariadelmarq seems to work!

PR #1789, opened 2018-10-23
Hack to not separate program names with commas in them into different cells

mariadelmarq:
... when exporting results of optimization to spreadsheet. Fixes Trello card: https://trello.com/c/jmPc8osD/1047-optimization-output-parsed-due-to-program-name-commas.

### DIFF
--- a/optima/results.py
+++ b/optima/results.py
@@ -584,14 +584,11 @@ class Resultset(object):
                 budget_change = (optimized_budget - baseline_budget)/baseline_budget
             if baseline_cov > 0:
                 cov_change = (optimized_cov - baseline_cov) / baseline_cov
-            outputstr += prog + sep + '%s' % sigfig(baseline_budget, sigfigs=sigfigs) + sep + \
-                         '%sperc' % sigfig(baseline_budget / total_baseline_budget, sigfigs=sigfigs) + sep + \
-                         '%s' % sigfig(optimized_budget, sigfigs=sigfigs) + sep + \
-                         '%sperc' % sigfig(optimized_budget / total_optimized_budget, sigfigs=sigfigs) + \
-                         sep + '%spercconditional' % sigfig(budget_change, sigfigs=sigfigs) + sep + \
-                         '%s' % sigfig(baseline_cov, sigfigs=sigfigs) + sep + \
-                         '%s' % sigfig(optimized_cov, sigfigs=sigfigs) + sep + \
-                         '%spercconditional' % sigfig(cov_change, sigfigs=sigfigs)
+            outputstr += prog + sep + str(baseline_budget) + sep + \
+                         '%fperc' % (baseline_budget/total_baseline_budget) + sep + str(optimized_budget) + sep + \
+                         '%fperc' % (optimized_budget/total_optimized_budget) + sep + \
+                         '%fpercconditional' % budget_change + sep + str(baseline_cov) + sep + \
+                         str(optimized_cov) + sep + '%fpercconditional' % cov_change
         outputstr += '\n'
         outputstr += sep.join(['Total', str(total_baseline_budget), '', str(total_optimized_budget), '', '',
                                '', '', ''])
--- a/optima/results.py
+++ b/optima/results.py
@@ -1096,7 +1096,7 @@ def exporttoexcel(filename=None, outdict=None):
         outlist = []
         for line in outstr.split('\n'):
             outlist.append([])
-            if '"' in line:
+            if '","' in line:
                 for cell in line.split('","'):  # Hack to not separate program names with commas in them
                     outlist[-1].append(str(cell))  # If unicode, doesn't work
             else:
--- a/optima/results.py
+++ b/optima/results.py
@@ -498,8 +498,12 @@ class Resultset(object):
                 else:                       data = self.main[mainkey].tot[ind][:]
                 outputstr += self.main[mainkey].name+sep+popkey+sep
                 for t in range(npts):
-                    if self.main[mainkey].ispercentage: outputstr += ('%s'+sep) % sigfig(data[t]*100, sigfigs=sigfigs)
-                    else:                           outputstr += ('%i'+sep) % data[t]
+                    if '(per 100 p.y.)' in self.main[mainkey].name:
+                        outputstr += ('%s' + sep) % sigfig(100 * data[t], sigfigs=sigfigs)
+                    elif self.main[mainkey].ispercentage:
+                        outputstr += ('%sperc'+sep) % sigfig(data[t], sigfigs=sigfigs)
+                    else:
+                        outputstr += ('%i'+sep) % data[t]
 
         if not comparisontab:
             # Handle budget and coverage
@@ -554,7 +558,7 @@ class Resultset(object):
         else:
             return outputstr
 
-    def comparebudgets(self, sep='\",\"'):
+    def comparebudgets(self, sep='\",\"', sigfigs=3):
         """ Make a separate sheet in the workbook with a comparison of budget and coverage for an optimization """
         outputstr = sep.join(['Programs', 'Baseline budget', '% baseline budget', 'Optimized budget',
                               '% optimized budget', '% budget change',
@@ -580,11 +584,14 @@ class Resultset(object):
                 budget_change = (optimized_budget - baseline_budget)/baseline_budget
             if baseline_cov > 0:
                 cov_change = (optimized_cov - baseline_cov) / baseline_cov
-            outputstr += prog + sep + \
-                         str(baseline_budget) + sep + '%fperc' % (baseline_budget/total_baseline_budget) + sep + \
-                         str(optimized_budget) + sep + '%fperc' % (optimized_budget/total_optimized_budget) + \
-                         sep + '%fpercconditional' % budget_change + sep + str(baseline_cov) + sep + \
-                         str(optimized_cov) + sep + '%fpercconditional' % cov_change
+            outputstr += prog + sep + '%s' % sigfig(baseline_budget, sigfigs=sigfigs) + sep + \
+                         '%sperc' % sigfig(baseline_budget / total_baseline_budget, sigfigs=sigfigs) + sep + \
+                         '%s' % sigfig(optimized_budget, sigfigs=sigfigs) + sep + \
+                         '%sperc' % sigfig(optimized_budget / total_optimized_budget, sigfigs=sigfigs) + \
+                         sep + '%spercconditional' % sigfig(budget_change, sigfigs=sigfigs) + sep + \
+                         '%s' % sigfig(baseline_cov, sigfigs=sigfigs) + sep + \
+                         '%s' % sigfig(optimized_cov, sigfigs=sigfigs) + sep + \
+                         '%spercconditional' % sigfig(cov_change, sigfigs=sigfigs)
         outputstr += '\n'
         outputstr += sep.join(['Total', str(total_baseline_budget), '', str(total_optimized_budget), '', '',
                                '', '', ''])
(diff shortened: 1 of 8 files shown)


mariadelmarq:
@cliffckerr @robynstuart @gchadder3 This branch formats the exported spreadsheet for optimization results according to the Application team's needs cc @sherriekelly . To make this the default from the FE, the parameter comparisontab should be set to True when exporting an optimization result (I didn't do it myself because the same function is used to export both calibration and optimization results and I didn't want to be too hacky). Can we request that this be reviewed, improvedif needed, and merged ASAP? Thanks!!!

optima/results.py
```diff
@@ -470,10 +470,10 @@ def assemble(key):
             self.other['only'+healthkey].tot =  process(allpeople[:,healthinds,:,:][:,:,:,indices].sum(axis=(1,2))) # Axis 1 is populations
 
         return None
-        
-        
-    def export(self, filename=None, folder=None, bypop=True, sep=',', ind=None, key=None, sigfigs=3, writetofile=True, asexcel=True, verbose=2):
-        ''' Method for exporting results to an Excel or CSV file '''
+
+    def export(self, filename=None, folder=None, bypop=True, sep=',', ind=None, key=None, sigfigs=3, writetofile=True,
+               comparisontab=False, asexcel=True, verbose=2):
```
cliffckerr: I think this logic should be rethought a bit...I don't see any reason to _not_ always include the comparison tab in addition to the other sheets for a Multiresultset but I also don't see why you would want it for a single Resultset, so my preference would be to change around the logic a bit and have comparisontab only apply to `Multiresultset` and just be true (not a kwarg).

optima/results.py
```diff
@@ -487,57 +487,62 @@ def export(self, filename=None, folder=None, bypop=True, sep=',', ind=None, key=
...
-            tvdata   = tvbudget['tvbudgets'][:]
-            outputstr += '\n\n\n'
-            outputstr += sep*2+'Time-varying' + '\n'
-            outputstr += sep+sep.join(['Year']+progkeys) + '\n'
-            for y,year in enumerate(tvyears): # Loop over years as rows
-                outputstr += sep+str(year)+sep+sep.join([str(val) for val in tvdata[:,y]]) + '\n' # Join together programs as columns
+                    if '(per 100 p.y.)' in self.main[mainkey].name:
+                        outputstr += ('%s' + sep) % sigfig(100 * data[t], sigfigs=sigfigs)
+                    elif self.main[mainkey].ispercentage:
+                        outputstr += ('%sperc'+sep) % sigfig(data[t], sigfigs=sigfigs)
```
cliffckerr: Maybe  `('%s%%'+sep)` would be better? (i.e. percent symbol rather than text)

optima/results.py
```diff
@@ -487,57 +487,62 @@ def export(self, filename=None, folder=None, bypop=True, sep=',', ind=None, key=
...
+                    else:
+                        outputstr += ('%i'+sep) % data[t]
+
+        if not comparisontab:
```
cliffckerr: If `comparisontab` is always true this will have to be rethought. Also, I would be concerned that this would break with trying to export results that don't have a budget (e.g. a Parscen).

optima/results.py
```diff
@@ -552,7 +557,47 @@ def export(self, filename=None, folder=None, bypop=True, sep=',', ind=None, key=
...
+            if baseline_budget <= 0.1 < optimized_budget:
+                budget_change = 1.0
+            elif baseline_budget > 0.1:
+                budget_change = (optimized_budget - baseline_budget)/baseline_budget
+            if baseline_cov <= 0.1 < optimized_cov:
+                cov_change = 1.0
+            elif baseline_cov > 0.1:
+                cov_change = (optimized_cov - baseline_cov) / baseline_cov
+            outputstr += prog + sep + str(baseline_budget) + sep + \
+                         '%fperc' % (baseline_budget/total_baseline_budget) + sep + str(optimized_budget) + sep + \
```
cliffckerr: See above about comment using %% instead of perc

optima/results.py
```diff
@@ -552,7 +557,47 @@ def export(self, filename=None, folder=None, bypop=True, sep=',', ind=None, key=
...
+                baseline_cov = baseline_cov[0]  # Only pull out the first element if it's an array/list
+            if baseline_cov is None: baseline_cov = 0  # Just reset
+            optimized_cov = self.coverages['Optimal'][prog]
+            if checktype(optimized_cov, 'arraylike'):
+                optimized_cov = optimized_cov[0]  # Only pull out the first element if it's an array/list
+            if optimized_cov is None: optimized_cov = 0  # Just reset
+
+            budget_change = 0
+            cov_change = 0
+            if baseline_budget <= 0.1 < optimized_budget:
```
cliffckerr: This logic seems a little convoluted to me, having two less-than operators in a single conditional...maybe rewrite, and also where does 0.1 come from?

optima/results.py
```diff
@@ -833,21 +874,34 @@ def diff(self, base=None):
                 resultsobj.pops[simkey] -= basepop
         
         return resultsdiff
-    
-    
-    def export(self, filename=None, folder=None, ind=None, writetofile=True, verbose=2, asexcel=True, **kwargs):
-        ''' A method to export each multiresult to a different sheet in Excel (or to a single large text file) '''
+
+    def export(self, filename=None, folder=None, ind=None, writetofile=True, verbose=2, asexcel=True,
+               comparisontab=False, **kwargs):
```
cliffckerr: See above about comparisontab just being true and not a kwarg

optima/results.py
```diff
@@ -833,21 +874,34 @@ def diff(self, base=None):
...
+        """ A method to export each multiresult to a different sheet in Excel (or to a single large text file) """
         
-        if asexcel: outputdict = odict()
-        else:       outputstr = ''
+        if asexcel:
+            outputdict = odict()
+        else:
+            outputstr = ''
```
cliffckerr: I assume this was done automatically by PyCharm..I think the original was easier to read though :)

optima/results.py
```diff
@@ -833,21 +874,34 @@ def diff(self, base=None):
...
         for key in self.keys:
-            thisoutput = Resultset.export(self, ind=ind, key=key, writetofile=False, **kwargs)
+            thisoutput = Resultset.export(self, ind=ind, key=key, writetofile=False, comparisontab=comparisontab,
+                                          **kwargs)
```
cliffckerr: As above re: kwarg

optima/results.py
```diff
@@ -833,21 +874,34 @@ def diff(self, base=None):
...
+
+        if comparisontab:
+            thisoutput = Resultset.comparebudgets(self)
+            if asexcel:
+                outputdict['Comparison'] = thisoutput
+            else:
+                outputstr += 'Comparison\n\n'
+                outputstr += thisoutput
+                outputstr += '\n' * 5
+
```
cliffckerr: Have you tested that this does still work when writing to CSV instead of Excel?

optima/results.py
```diff
@@ -1028,50 +1078,86 @@ def exporttoexcel(filename=None, outdict=None):
     workbook = Workbook(filename)
     
     for key,outstr in outdict.items():
-        worksheet = workbook.add_worksheet(sanitizefilename(key)) # A valid filename should also be a valid Excel key
+        if key == 'Optimal':
+            key = 'Optimized'  # Hack to replace optimal with optimized for sheet name
```
cliffckerr: Why is this being done? If we want to rename "Optimal" to "Optimized" gobally then we can, but I think we should do it consistently, otherwise it may cause problems and inconsistencies down the track.

optima/results.py
```diff
@@ -1028,50 +1078,86 @@ def exporttoexcel(filename=None, outdict=None):
...
-            for cell in line.split(','):
-                if cell=='tot': cell = 'Total' # Hack to replace internal key with something more user-friendly
-                outlist[-1].append(str(cell)) # If unicode, doesn't work
+            if '"' in line:
+                for cell in line.split('","'):  # Hack to not separate program names with commas in them
+                    outlist[-1].append(str(cell))  # If unicode, doesn't work
+            else:
+                for cell in line.split(','):
+                    if cell == 'tot': cell = 'Total'  # Hack to replace internal key with something more user-friendly
+                    outlist[-1].append(str(cell))  # If unicode, doesn't work
```
cliffckerr: I'm not sure this logic is going to be very robust...the `if '"' in line` seems too broad, e.g. won't this break if you have a quote in a program name? I think a safer approach would be to do sanitization on the program names before they get here and strip out any commas (and possibly any other invalid characters).

optima/results.py
```diff
@@ -1028,50 +1078,86 @@ def exporttoexcel(filename=None, outdict=None):
...
+                            try:
+                                thistxt = thistxt.replace("conditional", "")
+                                thistxt = float(thistxt)
+                                percentcell = True
+                                if thistxt > 0:
+                                    percentincrease = True
+                                elif thistxt < 0:
+                                    percentdecrease = True
+                            finally:
+                                pass
```
cliffckerr: Rather than do this with a series of try-catches, I'd just check directly, and I'd probably also make the labels more unique (since e.g. a program name could easily have 'perc' in it):
```
percstr = '!PERC' # define above
condstr = '!COND' # ditto
percentcell = percstr in thistxt # will be true if found, false otherwise, and don't need to define above
conditionalcell = condstr in thistxt
thistxt = thistxt.replace(percstr, '').replace(condstr, '')
try:
  thistxt = float(thistxt)
  numbercell = True
except:
  numbercell = False
```

mariadelmarq: Agreed, will update!

optima/results.py
mariadelmarq: @sherriekelly requested this specifically, maybe @cliffckerr and @robynstuart can chat with @sherriekelly to reach a consensus?

optima/results.py
mariadelmarq: perc was the label for indicating a cell was to be formatted as percent in excel. Updated to more unique label "!PERC"

optima/results.py
mariadelmarq: perc was the label for indicating a cell was to be formatted as percent in excel. Updated to more unique label "!PERC"

optima/results.py
mariadelmarq: @cliffckerr My hack only works for optimization results (baseline vs optimized), could you advice on the best way to make it the default behavior only whenever it is an optimization multiresult?

optima/results.py
mariadelmarq: @cliffckerr: See above, we would want this code to be executed whenever it is an optimization multiresult

optima/results.py
mariadelmarq: @cliffckerr: See above, we would want this code to be executed whenever it is an optimization multiresult

optima/results.py
mariadelmarq: Good point, it doesn't, so this code can only execute when writing to excel, as well as being an optimization multiresultset.

optima/results.py
mariadelmarq: Fixed :)

optima/results.py
mariadelmarq: @cliffckerr: See above, we would want this code to be executed whenever it is an optimization multiresult

optima/results.py
mariadelmarq: I added all of "," as a separator of cells only for program names (instead of just ,), but I agree it's hacky, not very robust, and could fail (if a program name contained "," for some reason...). Is there already a sanitizing code for strings/program names? If so, please feel free to change and commit on top of this.

optima/results.py
mariadelmarq: @cliffckerr: 0.1 was just a tolerance value, to not use == 0, which I know can be problematic. Please feel free to improve. Pycharm was complaining about writing the two separate conditionals, which is why I combined, but feel free to separate again.

PR #1791, opened 2018-10-24
Use P.name as filename when downloading project from FE

mariadelmarq:
Resolves Trello card: https://trello.com/c/MUZOa0jr/1049-retaining-project-names-when-downloading


mariadelmarq:
@cliffckerr : you're absolutely right, sorry about that! I got confused with the `<downloadProject>` and `<downloadPrjWithResults>` javascript functions in *project.js* taking in _name_ as a parameter, but not using it within the function!

PR #1792, opened 2018-10-24
Replace "current" with "most recent"

mariadelmarq:
Fixes Trello card: https://trello.com/c/GYhr6v1E/1043-relabel-optimization-edit-current-to-most-recent

### DIFF
--- a/client/source/js/modules/optimization/optimization-modal.html
+++ b/client/source/js/modules/optimization/optimization-modal.html
@@ -191,8 +191,8 @@
           border-top: 1px solid #999;
           border-bottom: 1px solid #999">
       <th style="vertical-align:bottom">Program</th>
-      <th style="width: 6em">Not less than<br>(% of current)</th>
-      <th style="width: 5em">Not more than<br>(% of current)</th>
+      <th style="width: 6em">Not less than<br>(% of most recent)</th>
+      <th style="width: 5em">Not more than<br>(% of most recent)</th>
     </tr>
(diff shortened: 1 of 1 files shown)


client/source/js/modules/optimization/optimization-modal.html
```diff
@@ -191,8 +191,8 @@ <h3>
           border-top: 1px solid #999;
           border-bottom: 1px solid #999">
       <th style="vertical-align:bottom">Program</th>
-      <th style="width: 6em">Not less than<br>(% of current)</th>
-      <th style="width: 5em">Not more than<br>(% of current)</th>
+      <th style="width: 6em">Not less than<br>(% of most recent)</th>
+      <th style="width: 5em">Not more than<br>(% of most recent)</th>
```
cliffckerr: I'm not sure about this change actually. In the _model_, it is interpreted as a percentage of spending at the starting year of the optimization. This is usually the "current" spending which also happens to be the "most recent" spending, but neither of these need be the case -- you can for example start an optimization in 2010, for example.

Thus I think "most recent" is confusing since it isn't necessarily true, and also obscures the fact that in every case where "most recent" is correct, so is "current". Thus I would suggest either leaving as-is, or changing to "Not less than (% of funding at start year)" or "Not less than (% of initial funding"). @robynstuart what do you think?

cliffckerr: replaced with different text, see slack

PR #1793, opened 2018-10-29
fix percentage export bug

cliffckerr:
Fixes long-standing bug that exported results (but not plots) show absolute values instead of percentages (which also meant that incidence was showing as per year instead of per 100 years).


mariadelmarq:
Redundant: https://github.com/optimamodel/optima/pull/1789

PR #1800, opened 2019-01-10
Python3 compatibility new

cliffckerr:
Almost working with Python 3...


optima/programs.py
```diff
@@ -572,17 +576,20 @@ def overlap_calc(indexes,target_depth):
...
                                 for j in range(indexes[-1]+1,len(thiscov)):
                                     accum += overlap_calc(indexes+[j],target_depth)
-                                return thiscov.values()[indexes[-1]]*accum
+                                output = list(thiscov.values())[indexes[-1]]*accum
+                                return output
                             else:
-                                return thiscov.values()[indexes[-1]]* max([delta.values()[x] for x in [0]],0)
+                                deltalist = list(delta.values())
+                                output = list(thiscov.values())[indexes[-1]]* max(deltalist[0][0],0)
+                                return output
```
cliffckerr: @robynstuart this was failing for mysterious reasons in python3, but how it _works_ was kind of mysterious to me...if you can check this change, that'd be much appreciated!

robynstuart: @cliffckerr it was definitely wrong before... I think we want
`output = thiscov.values()[indexes[-1]] * max(delta.values()[0], 0)`

NB, @RomeshA has introduced a cleaner implementation in Atomica - @RomeshA, do you think this is worth porting to HIV?

PR #1802, opened 2019-06-28
Python3 website

RomeshA:
Enable Python 3 compatibility for the FE. Should be backwards compatible but need to double check it's OK on athena with Python 2. Note that the python versions cannot be changed without migrating the database pickles (so this PR can be merged at any point but athena should not be changed to python 3 yet)

### DIFF
--- a/server/config.example.py
+++ b/server/config.example.py
@@ -15,10 +15,12 @@ SQLALCHEMY_TRACK_MODIFICATIONS = bool(os.getenv('SQLALCHEMY_TRACK_MODIFICATIONS'
 MATPLOTLIB_BACKEND = os.getenv('MATPLOTLIB_BACKEND',"agg")
 SERVER_PORT = int(os.getenv('PORT', 8080))
 WORKERS = int(os.getenv('WORKERS',min(math.ceil(multiprocessing.cpu_count()/2.0),8))) # By default use half the CPUs, up to a maximum of 8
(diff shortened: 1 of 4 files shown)


server/config.example.py
```diff
@@ -1,10 +1,24 @@
...
+CELERY_BROKER_URL = os.getenv('CELERY_BROKER_URL',REDIS_URL)
+CELERY_RESULT_BACKEND = os.getenv('CELERY_RESULT_BACKEND',REDIS_URL)
+CELERY_ACCEPT_CONTENT = os.getenv('CELERY_ACCEPT_CONTENT','pickle,json,msgpack,yaml').split(',') # Comma separated list
+SQLALCHEMY_TRACK_MODIFICATIONS = bool(os.getenv('SQLALCHEMY_TRACK_MODIFICATIONS',False))
+MATPLOTLIB_BACKEND = os.getenv('MATPLOTLIB_BACKEND',"agg")
+SERVER_PORT = int(os.getenv('PORT', 8080))
+WORKERS = int(os.getenv('WORKERS',min(math.ceil(multiprocessing.cpu_count()/2.0),8))) # By default use half the CPUs, up to a maximum of 8
+
+# GUNICORN VARIABLES
+bind = '0.0.0.0:%d' % (SERVER_PORT) # Bind to 0.0.0.0 so apollo can talk to athena. The port should generally be publicly firewalled
```
cliffckerr: @RomeshA It seems like this should be something from an environment variable or the systemctl script, not in the Optima HIV repo?

RomeshA: Let's just make it a default with an env variable override like the others. I think `0.0.0.0` would be fine under most circumstances. I'll also clarify the comment - it's not actually that specific to athena/apollo, more about the choice of `0.0.0.0` vs `127.0.0.1` depending on whether you want to be able to connect to it from other machines or not

PR #1816, opened 2019-10-31
Remove a potential divide by zero

Rowanmh:
... leading eventually to non-positive people error

### DIFF
--- a/optima/optimization.py
+++ b/optima/optimization.py
@@ -1383,7 +1383,8 @@ def minmoney(project=None, optim=None, tvec=None, verbose=None, maxtime=None, ma
                     trial_vec[ind] = 0 # Set to zero temporarily
                     new_prog = this_prog*refine_step
                     reallocate = this_prog*(1-refine_step)*refine_keep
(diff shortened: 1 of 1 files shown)


cliffckerr:
Actually @Rowanmh  looking at it more closely, if `trial_vec.sum()` is zero then there's no point doing this loop, so best just to terminate, no?

ckerr-IDM:
@Rowanmh I'm still curious about the case that resulted in this error. It seems like it should only happen if you're trying to optimize a budget of $0. The error could be more helpful, but I do think it's appropriate to throw an error in that case :) Do you remember what caused this originally?

Rowanmh:
@cliffk - this happened when running a money minimization with some aggressive but achievable targets near saturation

cliffckerr:
@Rowanmh Really? I don't understand how the entire vector is 0 though...my feeling is there's probably a deeper bug here. If you're able to reproduce reliably (setting the random seed), would be good to debug further...

cliffckerr:
Superseded by #1814

PR #1817, opened 2019-11-20
fix percentage export bug

cliffckerr:
Need to review if this is still an issue


cliffckerr:
Never mind, this was resolved in another commit

PR #1818, opened 2019-11-20
Prs 2019jul15 take2

cliffckerr:
Something was weird with the other PR... Also, this breaks something, so don't merge yet.


cliffckerr:
Probably safe to say this PR isn't needed after 5 years :)

PR #1819, opened 2019-11-23
Export the cascade proportions as well as numbers in portfolio output

Rowanmh:


cliffckerr:
@Rowanmh @sherriekelly @hussainazfar  I've made some modifications and now all tests pass for me. Do you want to check it or should I just merge?

PR #1823, opened 2020-01-13
Optimization fixes

RomeshA:
I'm not sure what's causing the optimization issues on Apollo but this PR makes some changes to the way the FE communicates with the BE in optimization. In particular

- The elapsed time should no longer get faster and faster with successive runs
- Optimizations can be cancelled

There's a chance that these changes modify/remove the code causing the issue and that the issue will be resolved automatically. If not, I'm hoping it'll be a little easier to isolate what's going wrong with the changes made here.

@cliffckerr (mainly) - cancelling tasks is accomplished via `stoppingfunc` but it's necessary to add some calls to `stoppingfunc` at various points in the code to achieve termination in a timely fashion. ASD has support for gracefully terminating (returning whatever it got up to) but as the FE shouldn't display partial results and cancelling takes effect immediately, it's equally fine to cancel via

```
    if stoppingfunc and stoppingfunc():
        raise op.CancelException
```

where `CancelException` is a new error class that is specifically caught in `run_task()`. I've dropped in a few of those checks in various places during `minoutcomes` (e.g. if you cancel a task, then it shouldn't try and start ASD for all of the ~11 runs or so only to have them each terminate after one iteration). So it would be great at some point if you could put a few of these checks into `minmoney` as well


optima/optimization.py
```diff
@@ -212,23 +212,22 @@ def constrainbudget(origbudget=None, budgetvec=None, totalbudget=None, budgetlim
 
     # Calculate the current total budget
     currenttotal = sum(constrainedbudget[:]) # WARNING, assumes it's an odict
-    scaleratio = totalbudget/float(currenttotal) # Calculate the ratio between the original budget and the supplied budget
```
RomeshA: @cliffckerr I only changed `optimization.py` to add `stoppingfunc` in various places - the other changes might accidentally be reverting updates due to the order in which they were committed on my branch - can you please double check which is the correct version?

RomeshA:
@cliffckerr OK this one should hopefully also be good to go now, I went through and removed the extra changes from `optimization.py` relative to the `develop` branch

PR #1826, opened 2020-02-18
Disable click to close and escape to close

RomeshA:
Prevents accidental data loss particularly on modals with many values


RomeshA:
@cliffckerr This one should be ready to go

PR #1869, opened 2023-01-31
Inf budget gets scaled up properly

kelvinburke:
For minmoney, the budget gets scaled up by setting totalbudget and letting constrainbudget scale it up. This previously multiplied the current budget by the necessary factor which left 0 budgets at 0.

Now, the budget gets scaled up by adding the difference from the current budget evenly between all programs. This means that for inf budget, each program gets inf budget / nprograms.

We could change that inf budget actually is infbudget*nprograms so that each program gets infbudget added?

This doesn't change that constraints are relative yet.

It seems to produce good results from the brief testing that I have done, but shouldn't produce worse results as this is only used when scaling up, not scaling down - and the minmoney optimization makes sure to try to scale each program down appropriately.

### DIFF
--- a/optima/optimization.py
+++ b/optima/optimization.py
@@ -254,7 +254,8 @@ def calcoptimindsoptimkeys(functionname, prognamelist=None, progset=None, optimi
     return optiminds, optimkeys
 
 
-def constrainbudget(origbudget=None, budgetvec=None, totalbudget=None, budgetlims=None, optiminds=None, optimkeys=None, tolerance=1e-2, overalltolerance=1.0, outputtype=None, verbose=2, tvsettings=None):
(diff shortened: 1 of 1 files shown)


Rowanmh:
In theory `infbudget` should be high enough that this should never make a difference and should always result in full saturation for all programs even if divided by the number of programs. I think it's safer to leave as is then (less chance of integer overflow say if we increase infbudget)

kelvinburke:
Agreed, at the moment a kind of worst case is that the infbudget = 1e10, is split between say 100 programs meaning each program only gets $100M. I guess also depends if we use a different currency which is worth less, say Indonesian Rupiah. Then 100M IDR is only US$10,000 approx. Perhaps we should increase infbudget to say 1e14? Noting that doubles which python uses have 53 bits of precision = 15 and a bit digits of precision. Although that shouldn't matter since we don't care about the ones column if we have inf budget.

PR #1891, opened 2023-03-22
Actsreg, actscas, actscom now only have insertive acts in them

kelvinburke:
So directionality in the Partnerships matrices is now observed. If both directions are put in the matrix then the ratio of the two directions in the matrix gives the ratio of the insertive acts.


kelvinburke:
Plus relhivbirth only applies to diagnosed mothers with HIV, not all PLHIV, and not those diagnosed by ANC

PR #1893, opened 2023-03-23
Actsreg, actscas, actscom now only have insertive acts in them

kelvinburke:
Finished version of https://github.com/optimamodel/optima/pull/1891

Adds a migration so that results using previous parsets are not changed - only changed a small amount by the PMTCT changes if you are using 2.12.0 or later. Only uses the new directional partnerships if you recreate the parset from the data - pars['actsreg'].insertiveonly is the attribute which determines whether the old or new behaviour.

Directionality in the Partnerships matrices is now observed. If both directions are put in the matrix then the ratio of the two directions in the matrix gives the ratio of the insertive acts.

Plus relhivbirth only applies to diagnosed mothers with HIV, not all PLHIV, and not those diagnosed by ANC


kelvinburke:
Already cherry-picked the commits into `updates-june-2022` which were merged into `EECA-training-2022` in d0fb0ebc21db77d03d56bea8bdb1e9dfee0907f1

PR #1900, opened 2023-07-04
Rework init and min-money2

kelvinburke:
Previously optima.__init__ imported the projects file as optima._project so that it wouldn't get confused with the class optima.Project.

But this causes errors when unpickling for some reason, so instead we import the projects file as optima.project and the class is either optima.Project or optima.project.Project

Similarly for all the other files


kelvinburke:
Cherry-picked the rework init changes into both EECA-training-2022 and updates-june-2022 and the minmoney-asd changes are in minmoney-asd branch

PR #1951, opened 2026-02-20
Rev 15

kelvinburke:
Small bug fixes

### DIFF
--- a/optima/optimization.py
+++ b/optima/optimization.py
@@ -8,7 +8,7 @@ from optima import OptimaException, Link, Multiresultset, ICER, asd, getresults 
 from optima import printv, dcp, odict, findinds, today, getdate, uuid, objrepr, promotetoarray, findnearest, sanitize, \
     inclusiverange, sigfig, compareversions, cpu_count # Utilities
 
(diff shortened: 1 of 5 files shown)


optima/optimization.py
```diff
@@ -673,7 +673,7 @@ def outcomecalc(budgetvec=None, which=None, project=None, parsetname=None, progs
     tvec       = project.settings.maketvec(end=objectives['end'])
     # initpeople = None # Not anymore! # WARNING, unfortunately initpeople is still causing mismatches -- turning off for now despite the large (2.5x) performance penalty
     if initpeople is None: startind = None
-    elif startind is None: startind = findnearest(tvec, objectives['start']) # Assume initpeople is from the start of the optimization
+    elif startind is None: raise Exception('initpeople is set but startind is not! Do not know when to start from!') # startind = findnearest(tvec, objectives['start']) # Assume initpeople is from the start of the optimization
```
kelvinburke: @Rowanmh This was assuming that startind was at `objectives['start']` but really should have been `objectives['start'] - 1` and now could be `objectives['base'] - 1`. It was an unsafe assumption anyway so removed.

Just need to run some outcome optimizations to make sure this doesn't throw an error now

kelvinburke: Don't think any errors

